PR #143, opened 2021-10-22
Bugfix/asset types

fbellgr:
# PR details

- [ x] Bug fix
- [x ] Enhanced metadata

- Asset.definition was missing typeMapping entry for .ai
- Method Asset._readExtendedFileFromFS was being called with wrong parameters order
- Changed the filter in Asset to use type ids instead of type names, which solves the 10k limit at the source.

Not really. I only have a note about the way the code was supposed to overcome the 10k limit before I changed the filter.
It did not allow me to retrieve more than 10k, hence this pull request.
But I have an app that use this API in a similar manner, so I knew there was no 10k limit when using type ids.

Maybe you would like to rewrite this part of the code since the first 'if' should never be true anymore?

`lib\metadataTypes\Asset.js:231`

```
if (response.body.message && response.body.message.includes('all shards failed')) {
    // When running certain filters, there is a limit of 10k on ElastiCache.
    // Since we sort by ID, we can get the last ID then run new requests from there
    options.json.query = {
        leftOperand: {
            property: 'assetType.id',
            simpleOperator: 'in',
            value: subtypeIds,
        },
        logicalOperator: 'AND',
        rightOperand: {
            property: 'id',
            simpleOperator: 'greaterThan',
            value: items[items.length - 1].id,
        },
    };
    lastPage = 0;
    moreResults = true;
} else if (response.body.page * response.body.pageSize < response.body.count) {
    moreResults = true;
    lastPage = Number(response.body.page);
} else {
    moreResults = false;
}
```

## Checklist

- [ x] I have performed a self-review of my own code

### DIFF
--- a/lib/metadataTypes/Asset.js
+++ b/lib/metadataTypes/Asset.js
@@ -162,6 +162,7 @@ class Asset extends MetadataType {
         } else {
             Util.logger.info(`- Caching Subtype: ${subType}`);
         }
+        const subtypeIds = subTypeArray?.map(subTypeItemName => Asset.definition.typeMapping[subTypeItemName]);
         const uri = 'asset/v1/content/assets/';
         const options = {
             uri: uri + 'query',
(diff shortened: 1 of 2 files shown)


JoernBerkefeld:
had to update the target to include it in the upcoming 3.1.0 release

JoernBerkefeld (approved):
good ideas! i am going to hold off on removing the extra "if" as you suggested until having had the chance to review that but happy to discuss it further.

FYI @DougMidgley

JoernBerkefeld:
added this change to the 3.1.0 changelog in #develop

Issue #166, opened 2022-01-11
[FEATURE] Move SQL formatting to Prettier Plugin


DougMidgley:
**Is your feature request related to a problem? Please describe.**
Currently SQL parsing for lib internal and for vscode are potentially different

**Describe the solution you'd like**
Use Prettier Plugin for SQL to format, rather than another lib to standardize parsing.

**Describe alternatives you've considered**
Keep as is

**Additional context**
This will ensure both vscode and internal processing produce the same results

PR #167, opened 2022-01-11
[Feature] #166 sql via prettier

DougMidgley:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [X] Add something to the core
- [ ] Other, please explain:

Standardized SQL formatting via Prettier

Think about potential cross-impacts
## Checklist

- [ X] I have performed a self-review of my own code
- [ X] I have commented my code, particularly in hard-to-understand areas
- [ X] ESLint & Prettier are not outputting errors or warnings
- [ X] README.md updated (if applicable)
- [ X] CHANGELOG.md updated
- [ X] ran `npm run docs` to update developer docs


JoernBerkefeld (changes requested):
lets merge the sdk PR #109 first

JoernBerkefeld:
also, does this change how sql is formatted at the moment?

JoernBerkefeld:
does prettier support the auto-uppercasing?

JoernBerkefeld:
hm https://www.npmjs.com/package/prettier-plugin-sql
"This plugin adds support for a lot of sql files through sql-formatter or node-sql-parser."
sooo it's just a wrapper and switches our processing form https://www.npmjs.com/package/sql-formatter-plus to https://www.npmjs.com/package/sql-formatter

I believe the reason for using the plus version was the existence of the vscode plugin for the plus version but not for the normal version. 

- [ ] what did your tests show on changed formatting?
- [ ] does this allow setting sql-formatter configs in the prettier config?

JoernBerkefeld:
i noticed another issue: 
this will require an overwrite of the settings.json during the upgrade. that action defaults to No at the moment to prevent accidental overwrites. We need to ensure we bug users enough to make them actually overwrite that (or at least merge the new changes in)

DougMidgley:
- Minor spacing issues are changed, otherwise constant
- Yes, there is minor configs (same with existing) but they can be set in prettier config
- Same outcomes as existing for template {{{ workaround
- If we need a settings update, we can always keep this PR until v4. No rush, just a simplification which i think makes sense to ensure the UI and server formatting match well

JoernBerkefeld:
I would assume we can fix the spacing difference with a config change though?
for the optional override I created feature request #172 which should be a prerequisite for closing this PR

- [ ] close #172

JoernBerkefeld (changes requested):
close #172 first

DougMidgley:
> @DougMidgley - lets push this into 4.0.0 - we only postponed it before because we didn't want to release a new major version but now.... :)

Indeed, downmerge done :) ready for review

JoernBerkefeld (approved):

JoernBerkefeld (changes requested):

JoernBerkefeld:
when #172 / #274 is merged I would ask that its changes are merged into this PR first and that the _there existing_ boilerplate/forcedUpdates.json is extended from:
image
to

```json
[
    {
        "version": "4.0.0",
        "files": [".vscode/settings.json", ".vscode/extensions.json", ".prettierrc"]
    },
    {
        "version": "3.1.0",
        "files": [".vscode/settings.json"]
    }
]
```

.prettierrc
```diff
@@ -21,6 +21,14 @@
...
                 "tabWidth": 2
             }
+        },
+        {
+            "files": "*.sql",
+            "options": {
+                "formatter": "sql-formatter",
+                "language": "tsql",
+                "uppercase": true
+            }
```
JoernBerkefeld, line 31: same change needs to made to the .prettierrc in /boilerplate/files/.prettierrc

JoernBerkefeld (changes requested):

.prettierrc
DougMidgley: updated

JoernBerkefeld (approved):

Issue #159, opened 2022-01-06
[FEATURE] Support Mobile Keywords


DougMidgley:
**Is your feature request related to a problem? Please describe.**
Mobile Keywords are supported in the Legacy Endpoints and these are often deployed in templated fashion across multiple business units

**Describe the solution you'd like**
Support for Retrieve, Deploy and Template of Keywords

**Describe alternatives you've considered**
Doing Manualls

**Additional context**

PR #174, opened 2022-01-13
[Feature]#159 mobilekeywords

DougMidgley:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [X] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [ ] Other, please explain:

Support for Mobile Code and Mobile Keywords

General review

## Checklist

- [-] CHANGELOG.md updated
- [X] ran `npm run docs` to update developer docs


lib/metadataTypes/MetadataType.js
```diff
@@ -1254,11 +1254,11 @@ class MetadataType {
     }
     /**
      *
-     * @param {Object} ex response payload from REST API
-     * @returns {void}
+     * @param {exception} ex response payload from REST API
```
JoernBerkefeld, line 1257: is that a valid jsdocs type?

lib/metadataTypes/MobileKeyword.js
```diff
@@ -0,0 +1,122 @@
...
+     * @returns {Promise<Object>} Promise of metadata
+     */
+    static async retrieveAsTemplate(templateDir, name, templateVariables) {
+        return super.retrieveREST(
+            templateDir,
+            `/legacy/v1/beta/mobile/keyword/?view=simple&$where=keyword%20eq%20%27${name}%27%20`,
+            null,
+            templateVariables
+        );
+        // TODO: Decide if we want to keep default handling (and move other types) or keep unique parsing?
```
JoernBerkefeld, line 43: lets have a separate clean-up task for consolidating the typed rt into one generic one and keep this one here - I'm assuming you would then uncomment the following lines?

JoernBerkefeld (changes requested):
please **update the draft release 3.2.0** when this is merged

lib/metadataTypes/MetadataType.js
DougMidgley: good point :)

lib/metadataTypes/MobileKeyword.js
DougMidgley: so i tried something a bit half way

lib/util/util.js
```diff
@@ -764,6 +764,26 @@ const Util = {
         const options = { stdio: [0, 1, 2] };
         return child_process.execSync(cmd + ' ' + args.join(' '), options);
     },
+    /**
+     * helper to standardize retrieveTemplate result checks
+     * @param {Array} results array of metadata
+     * @param {string} keyToSearch the field which contains the searched value
+     * @param {string} searchValue the value which is being looked for
+     * @returns {Object} metadata to be used in building template
```
JoernBerkefeld, line 772: 1st Param and returns could be typed better. We have definition for MetadataTypeItem

lib/util/util.js
```diff
@@ -764,6 +764,26 @@ const Util = {
         const options = { stdio: [0, 1, 2] };
         return child_process.execSync(cmd + ' ' + args.join(' '), options);
     },
+    /**
+     * helper to standardize retrieveTemplate result checks
```
JoernBerkefeld, line 768: Unclear what this does

JoernBerkefeld (changes requested):

lib/metadataTypes/MobileKeyword.js
JoernBerkefeld: lets have a call for this... tried understanding what you wrote there but failed

lib/util/util.js
DougMidgley: basically idea is to check that exactly 1 item is returned, then use it in other RT methods.

lib/util/util.js
DougMidgley: Still a bit new to me to use the types like this :) let me know if it works as this

JoernBerkefeld: i would usually use MetadataTypeItem[] here instead tbh but I'm not sure it works because of our missing imports. maybe we need to move the declaration into util.js
let me play with it...

DougMidgley: Cannot be covered right now, to be reassessed later with typescript implementation

Issue #175, opened 2022-01-13
[TASK] Run ESLINT as part of Code Analysis


DougMidgley:
When a PR is opened, both CodeQL and ESLINT should be run. Currently only CodeQL is run
Additionally, Commit & Branch formats should be checked on commit

PR #176, opened 2022-01-13
[Task] #175 es lint and hooks

DougMidgley:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [X] Other, please explain:

ESLINT validation on PR and Commit Checks

Test locally

## Checklist

- [-] CHANGELOG.md updated
- [-] ran `npm run docs` to update developer docs


JoernBerkefeld:
could you please update the draft release 3.2.0?
https://github.com/Accenture/sfmc-devtools/releases/edit/untagged-1ed1364d7f9c78e674b9

.husky/post-checkout
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+git config commit.template .git/templatemessage
+TICKETID=`git reflog | awk 'NR==1{ print $8; exit }' | LC_ALL=en_US.utf8 grep -oP '\d{3,5}'`
+echo "[POST_CHECKOUT] Setting template commit to #$TICKETID"
+echo "#$TICKETID " > ".git/templatemessage"
```
JoernBerkefeld, line 7: could you explain what that does? :)

package.json
```diff
@@ -52,8 +53,8 @@
         "sql-formatter-plus": "1.3.6",
         "toposort": "2.0.2",
         "update-notifier-git": "5.0.3",
-        "winston": "3.3.4",
-        "yargs": "17.3.0"
+        "winston": "3.4.0",
+        "yargs": "17.3.1"
```
JoernBerkefeld, line 57: any reason this was updated? dont see a note in the commits for it. if we don't warn about package changes, developers might not run `npm install` again, leading to issues.

package.json
```diff
@@ -64,7 +65,7 @@
         "eslint-plugin-prettier": "4.0.0",
         "husky": "7.0.4",
         "jsdoc-to-markdown": "7.1.0",
-        "lint-staged": "12.1.5",
+        "lint-staged": "12.1.7",
```
JoernBerkefeld, line 68: any reason this was updated? dont see a note in the commits for it. if we don't warn about package changes, developers might not run `npm install` again, leading to issues.

.husky/commit-msg
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
```
JoernBerkefeld, line 1: again, quick walk-through please

package.json
```diff
@@ -30,7 +30,8 @@
         "build": "run-s lint docs",
         "debug": "node --nolazy --inspect-brk=9229 lib/index.js",
         "docs": "jsdoc2md --files lib/**/*.js > docs/dist/documentation.md",
-        "lint": "eslint --fix lib/**/*.js && eslint --fix test/**/*.js && prettier --write lib/**/*.js",
+        "lint:fix": "eslint --fix lib/**/*.js && eslint --fix test/**/*.js && prettier --write lib/**/*.js",
+        "lint": "eslint lib/**/*.js && eslint test/**/*.js",
```
JoernBerkefeld, line 34: why the change? for the commit hook?

JoernBerkefeld (changes requested):
changelog needs updating once this is approved

package.json
DougMidgley: because lint shouldnt fix, it should lint only. This way we can run it in a check and not just hide issues

.husky/post-checkout
DougMidgley: when you checkeout a branch it takes the branch name and tries to get the ticket id. this then defaults the start of the commit message to #NNNN to help improve compliance in commit messages

.husky/commit-msg
DougMidgley: basically, this checks the default commit message (other comment) and checks that the commit message matches the correct format. If it is a merge, this is skipped to avoid problems

.husky/post-checkout
JoernBerkefeld: Fallback?
Is this executed Locally?

.husky/commit-msg
JoernBerkefeld: Fallback?

DougMidgley: meaning?

.husky/post-checkout
DougMidgley: its executed on any machine, local or not. In a workflow for example it will also execute, unless overriden (which you can also do). What would you look for in a fallback?

JoernBerkefeld: when you say "defaults" the commit message - does that impact what people see in their git clients, like sourcetree or gitfork? or where is this "defaulted"?

if it doesn't find an id (by the way, what is it looking for exactly? I cant figure out that grep logic...) it just prints `"# "` into .git/templatemessage?

JoernBerkefeld: never mind - im seeing this happens upon stage in gitfork, at the latest upon clicking into the commit message field - unless the field already has content in it

.husky/commit-msg
JoernBerkefeld: i mean what happens if the default commit message created in `post-checkout` is empty?

JoernBerkefeld: also, im seeing a different error message than the one you wrote?
image

.husky/post-checkout
JoernBerkefeld: tested this some more: if it does not find an id it seems to fail on line 5 and subsequently never update .git/templatemessage again. to test that, I first checked out your branch (175) and played around with the auto-completion. Then, I created a branch off of `develop`, switched to and merged your code in, afterwards I made a change and checked again. It would still autocomplete "#175 ".

same happened when i renamed your 175 branch to something without an id which i tried before the branch off of develop.

- [ ] add fallback for branches that don't have ids in them (e.g. created by external developers in their forked repos)

.husky/commit-msg
JoernBerkefeld: - [ ] check why the error message would talk about no matching files rather than the wrong commit message (tried committing something without "#123" upfront

JoernBerkefeld (changes requested):

.husky/commit-msg
DougMidgley: i pushed some changes but couldnt get this error. online search suggests running npm i again

JoernBerkefeld (approved):
all good now. merging

PR #218, opened 2022-03-27
#213 ensure docs/roles file is not overwritten during role deploy

### LINKED ISSUE #213: [BUG] deploy `role` overwrites docs/roles file

### Current Behavior

when you deploy 1 or multiple roles, the generic deploy method runs document for any types specified in the config. however, while that's good for multi-file documentation, it breaks for `role` (and probably `accountUser` as well), which gets documented into a single MD file per instance, rather than one MD file per record.

image

### Expected Behavior

deploy executes document only for multi-file documentation types 

### Steps To Reproduce

1. deploy one role
2. check the corresponding docs/roles/*.md file for your instance / credential file

### Version

3.1.3

### Environment

- OS:
- Node:
- npm:

### Participation

### Additional comments

_No response_

JoernBerkefeld:
# PR details

- [x] Bug fix

fix #213  

## Checklist

- [-] README.md updated (if applicable)
- [n/a] CHANGELOG.md updated

### DIFF
--- a/lib/metadataTypes/definitions/Role.definition.js
+++ b/lib/metadataTypes/definitions/Role.definition.js
@@ -13,6 +13,7 @@ module.exports = {
         'SYS_DEF_SECURITYADMIN',
         'SYS_DEF_VIEWER',
     ],
+    documentInOneFile: true,
     hasExtended: true,
     idField: 'ObjectID',
     keyField: 'CustomerKey',
--- a/lib/metadataTypes/MetadataType.js
+++ b/lib/metadataTypes/MetadataType.js
@@ -121,8 +121,12 @@ class MetadataType {
         const upsertResults = await this.upsert(metadata, deployDir, buObject);
         await this.postDeployTasks(upsertResults, metadata);
         const savedMetadata = await this.saveResults(upsertResults, retrieveDir, null);
-        if (this.properties.metaDataTypes.documentOnRetrieve.includes(this.definition.type)) {
+        if (
+            this.properties.metaDataTypes.documentOnRetrieve.includes(this.definition.type) &&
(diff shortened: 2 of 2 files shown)


lib/metadataTypes/definitions/Role.definition.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
         'SYS_DEF_SECURITYADMIN',
         'SYS_DEF_VIEWER',
     ],
+    documentInOneFile: true,
```
DougMidgley, line 16: I don't 100% agree with this approach. I actually wanna see document removed from deploy or retrieve. It should be its own thing. Then at CLI level it can be called after if need be on either. It will make the retrieve and deploy more "componentized", not to mention it would probably avoid the need for this modification since it would just take whatever is in the retrieve dir (which should be updated by whatever came from deploy)

DougMidgley (changes requested):

lib/metadataTypes/definitions/Role.definition.js
JoernBerkefeld: true - assuming the auto-documentation is a problem. 
on the other hand we used to see updates to our retrieve folder as a feature. and docs are created based on the retrieve folder. by disabling this we lead to difference between whats in retrieve and whats in docs.

it's 2 lines of code that can help us move forward with a fix of existing functionality instead of changing it.


I haven't seen a ticket nor a discussion prior to this bugfix about this, buddy. 
If we want to reduce whats done to docs/retrieve, we can add a new ticket for that and handle it as a feature change.

**Disputing a bugfix over a functional discussion seems wrong though.**

DougMidgley (approved):

PR #223, opened 2022-03-28
#222 improved debug output for callouts

### LINKED ISSUE #222: [TASK] get improved debug output for callout errors
Currently for some API issues like timeouts the error stack only includes one line.
With a bit of a workaround one can get the full stack trace though (compare top & bottom part - Error vs Trace):
image

idea is to always have this improved output in the log and offer it also to CLI if "--debug" was used in the command

JoernBerkefeld:
# PR details

- [x] Other, please explain:

fixes #222 

In certain cases we wanted a stack trace in our logs our CLI output. this was so far using Error.stack but as depicted in the issue, that's weirdly not always the best way. Instead, creating a NEW error and using that error's stack tends to do the trick of showing the full stack.

## Checklist

- [n/a] CHANGELOG.md updated

### DIFF
--- a/lib/Builder.js
+++ b/lib/Builder.js
@@ -91,9 +91,10 @@ class Builder {
             }
         } catch (ex) {
             Util.logger.error('mcdev.buildDefinition:' + ex.message);
-            Util.logger.debug(ex.stack);
+            const stack = new Error().stack;
+            Util.logger.debug(ex.message + ' ' + stack);
             if (Util.logger.level === 'debug') {
-                console.log(ex.stack);
+                console.log(ex.message, stack);
             }
         }
         return this.metadata;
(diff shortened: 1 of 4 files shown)


lib/Builder.js
```diff
@@ -91,9 +91,10 @@ class Builder {
             }
         } catch (ex) {
             Util.logger.error('mcdev.buildDefinition:' + ex.message);
-            Util.logger.debug(ex.stack);
+            const stack = new Error().stack;
```
DougMidgley, line 94: how does this even work? wont stack of a new Error always be null? or does it kind of know where it is?

JoernBerkefeld: nope it wont.
quoting the PR text:
"In certain cases we wanted a stack trace in our logs our CLI output. this was so far using Error.stack but as depicted in the issue, that's weirdly not always the best way. Instead, creating a NEW error and using that error's stack tends to do the trick of showing the full stack."

JoernBerkefeld: for more details please check the bug ticket where I supplied screenshots: #222

JoernBerkefeld:
as discussed via call, i moved the logic into a method inside of Util.logger

DougMidgley (approved):

PR #250, opened 2022-04-03
Task/94 replace update notifier git

DougMidgley:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [X] Other, please explain: minor changes to code

- Replaced update-notifier-git with update-notifier
- enforced correct usage of async instead of promises
- only request update on minor or major release (skip patch)

general check

## Checklist

- [X] CHANGELOG.md updated


package-lock.json
```diff
@@ -24,7 +24,7 @@
                 "simple-git": "3.5.0",
                 "sql-formatter-plus": "1.3.6",
                 "toposort": "2.0.2",
-                "update-notifier-git": "5.0.3",
+                "update-notifier": "5.1.0",
```
JoernBerkefeld, line 27: could you run a npm prune again to ensure we only have changes from this branch in here?

lib/util/util.js
```diff
@@ -119,40 +119,40 @@ const Util = {
      *
      * @param {object} properties javascript object in .mcdevrc.json
      * @param {boolean} [silent] set to true for internal use w/o cli output
-     * @returns {boolean|String[]} file structure ok OR list of fields to be fixed
+     * @returns {Promise<boolean|String[]>} file structure ok OR list of fields to be fixed
      */
-    checkProperties: function (properties, silent) {
-        if (!fs.existsSync(Util.configFileName) || !properties) {
+    checkProperties: async function (properties, silent) {
+        if (!(await fs.pathExists(Util.configFileName)) || !properties) {
```
JoernBerkefeld, line 125: why switch from exists to pathExists?
i got the sync/async part but that's a separate thing here

lib/util/util.js
```diff
@@ -119,40 +119,40 @@ const Util = {
...
             Util.logger.error(`\nCould not find ${Util.configFileName} in ${process.cwd()}.`);
             Util.logger.error(`Run 'mcdev init' to initialize your project.\n`);
             return false;
         }
-        if (!fs.existsSync(Util.authFileName) || !properties) {
+        if (!(await fs.pathExists(Util.authFileName)) || !properties) {
```
JoernBerkefeld, line 130: see above

lib/util/util.js
JoernBerkefeld: https://stackoverflow.com/questions/56488179/what-is-the-difference-between-fsextra-pathexists-and-fsextra-exists

lib/util/util.js
```diff
@@ -119,40 +119,40 @@ const Util = {
...
             return false;
         }
 
-        // check if user is running older mcdev version than whats saved to the config
-        if (properties.version && semver.gt(properties.version, packageJsonMcdev.version)) {
-            // dont run this for Catalyst to MC DevTools migration
+        // check if user is running older (ignores patches) mcdev version than whats saved to the config
+        if (
+            properties.version &&
+            ![null, 'patch'].includes(semver.diff(properties.version, packageJsonMcdev.version))
```
JoernBerkefeld, line 139: ok this is doing the limitation on patch which is cool. but we also need the gt() together with this (&&) to check if MY APP needs upgrading because the config is newer.

lib/util/util.js
```diff
@@ -284,7 +284,10 @@ const Util = {
             }
         }
         // check if project config version is outdated compared to user's mcdev version
-        if (!properties.version || semver.gt(packageJsonMcdev.version, properties.version)) {
+        if (
+            !properties.version ||
+            ![null, 'patch'].includes(semver.diff(packageJsonMcdev.version, properties.version))
+        ) {
```
JoernBerkefeld, line 290: like above, here we need the old solution combined with the new to check if the project's config file (in git) is outdated compared to my app's version.
yes, one could argue that here one check is sufficient cause otherwise we'd have exited earlier but that's not intuitive enough while reading the code

JoernBerkefeld (changes requested):

lib/util/util.js
JoernBerkefeld, line 139:
```suggestion
            properties.version &&
            ![null, 'patch'].includes(semver.diff(properties.version, packageJsonMcdev.version)) && semver.gt(properties.version, packageJsonMcdev.version)
```

lib/util/util.js
```diff
@@ -284,7 +284,10 @@ const Util = {
             }
         }
         // check if project config version is outdated compared to user's mcdev version
-        if (!properties.version || semver.gt(packageJsonMcdev.version, properties.version)) {
+        if (
+            !properties.version ||
+            ![null, 'patch'].includes(semver.diff(packageJsonMcdev.version, properties.version))
```
JoernBerkefeld, line 289:
```suggestion
            !properties.version ||
            (![null, 'patch'].includes(semver.diff(packageJsonMcdev.version, properties.version)) && semver.gt(packageJsonMcdev.version, properties.version) )
```

lib/util/util.js
```diff
@@ -119,40 +119,40 @@ const Util = {
...
+        ) {
             Util.logger.error(
                 `Your Accenture SFMC DevTools version ${packageJsonMcdev.version} is lower than your project's config version ${properties.version}`
             );
-            const questions = [
+            const responses = await inquirer.prompt([
                 {
                     type: 'confirm',
                     name: 'runUpgradeNow',
                     message: `Do you want to run 'npm update -g mcdev' now? This may take a few minutes.`,
```
JoernBerkefeld, line 148:
```suggestion
                    message: `Do you want to run 'npm update -g mcdev@${properties.version}' now? This may take a few minutes.`,
```

lib/util/util.js
DougMidgley: resolved in discussion

lib/util/util.js
DougMidgley: resolved in discussion

package-lock.json
```diff
@@ -1255,9 +1255,9 @@
             }
         },
         "node_modules/cliui/node_modules/ansi-regex": {
-            "version": "4.1.0",
```
JoernBerkefeld, line 1372: do we need this update? now this file is blocking the merge.

JoernBerkefeld (approved):

PR #262, opened 2022-04-09
Bugfix/259 smaller issues around create delta pkg and deploy

JoernBerkefeld:
# PR details

- [ ] Documentation update
- [x] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [ ] Other, please explain:

...

...

## Checklist

- [ ] CHANGELOG.md updated
- [x] ran `npm run docs` to update developer docs

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -741,11 +741,11 @@ class Mcdev {
 
     /**
      * Retrieve a specific metadata file and templatise.
-     * @param {String} businessUnit references credentials from properties.json
-     * @param {String} selectedType supported metadata type
-     * @param {String} name name of the metadata
-     * @param {String} market market which should be used to revert template
-     * @returns {Promise<Util.MultiMetadataTypeList>} -
+     * @param {string} businessUnit references credentials from properties.json
+     * @param {string} selectedType supported metadata type
+     * @param {string} name name of the metadata
+     * @param {string} market market which should be used to revert template
+     * @returns {Promise<{metadata:Util.MetadataTypeItem,type:string}>} -
      */
     static async retrieveAsTemplate(businessUnit, selectedType, name, market) {
         Util.logger.info('mcdev:: Retrieve as Template');
--- a/lib/index.js
+++ b/lib/index.js
@@ -155,12 +155,26 @@ class Mcdev {
                             buTypeDelta[bu][type].join(','),
                             sourceMarket
                         );
+                        console.log('type', type);
                         console.log('result', result);
+                        console.log('result.dataExtension', result.dataExtension);
+                        console.log('keys1', result[type]);
 
                         // ensure we have the right key for bd/bdb that matches the name used for rt
                         const keys = result[type]
-                            .map((item) => item[MetadataTypeInfo[type].definition.keyField])
+                            .map((item) => {
+                                console.log(
+                                    'MetadataTypeInfo[type].definition.keyField',
+                                    MetadataTypeInfo[type].definition.keyField
+                                );
+                                console.log(
+                                    'item[MetadataTypeInfo[type].definition.keyField]',
+                                    item[MetadataTypeInfo[type].definition.keyField]
+                                );
+                                return item[MetadataTypeInfo[type].definition.keyField];
+                            })
                             .filter((item) => !!item);
+                        console.log('keys2', keys);
                         if (keys.length) {
                             if (!typeDelta[type]) {
                                 // init array
--- a/lib/index.js
+++ b/lib/index.js
@@ -155,26 +155,10 @@ class Mcdev {
                             buTypeDelta[bu][type].join(','),
                             sourceMarket
                         );
-                        console.log('type', type);
-                        console.log('result', result);
-                        console.log('result[type]', result[type]);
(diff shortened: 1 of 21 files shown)


lib/util/util.js
```diff
@@ -50,6 +50,31 @@ const Util = {
...
+    /**
+     * helper to quickly ensure we are working with a flat Array of items
+     * @param {MetadataTypeMap|MetadataTypeItem[]} metadata map or array of metadata
+     * @returns {MetadataTypeItem[]} Array of metadata items
+     */
+    getMetadataMapToList(metadata) {
+        if (metadata && Array.isArray(metadata)) {
+            return metadata;
+        } else {
+            return Object.keys(metadata).map((item) => metadata[item]);
```
DougMidgley, line 62: I think you mean to use: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values
Also, I would wonder if this is really the right approach? We should always know what we are returning, and if this is needed it shows there is bugs which need fixing

lib/index.js
```diff
@@ -121,6 +121,7 @@ class Mcdev {
                 // Other map only with type -> file (for buildDefinitionBulk)
                 const buTypeDelta = {};
                 const typeDelta = {};
+                let deltaCounter = 0;
```
DougMidgley, line 124: not loving increasing logic creep in what is the index.js

lib/metadataTypes/Asset.js
```diff
@@ -101,7 +101,7 @@ class Asset extends MetadataType {
         }
         Util.logger.info(`Downloaded: ${this.definition.type} (${Object.keys(metadata).length})`);
 
-        return { metadata: metadata, type: this.definition.type };
+        return { metadata: metadata[Object.keys(metadata)[0]], type: this.definition.type };
```
DougMidgley, line 104: Also wondering if we shouldnt be doing this earlier? 
```suggestion
        return { metadata: Object.values(metadata)[0], type: this.definition.type };
```

lib/util/util.js
```diff
@@ -50,6 +50,31 @@ const Util = {
...
+        }
+    },
+    /**
+     * helper to turn an array or object into a single item
+     * @param {Object.<{metadata:MetadataTypeMap,type:string}>} obj map or array of metadata
+     * @returns {Object.<{metadata:MetadataTypeItem,type:string}>} item list reduced to single item
+     */
+    reduceMetadataMap(obj) {
```
DougMidgley, line 70: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values
also, i wonder if the name like "extractSingleObject" wouldnt be better? reduce has a specific meaning in js, so may be better to avoid, since it isnt reducing like you might expect

lib/metadataTypes/Script.js
```diff
@@ -53,15 +53,17 @@ class Script extends MetadataType {
...
      */
     static async retrieveAsTemplate(templateDir, name, templateVariables) {
         await File.initPrettier('ssjs');
-        return super.retrieveREST(
-            templateDir,
-            '/automation/v1/scripts/?$filter=name%20eq%20' + encodeURIComponent(name),
-            null,
-            templateVariables
+        return Util.reduceMetadataMap(
+            await super.retrieveREST(
```
DougMidgley, line 61: open point for discussion if retrieveREST wouldnt be the better place to put this? we could conditionally change result based on template.

DougMidgley (changes requested):

lib/index.js
JoernBerkefeld: Open to ideas on how to solve this more efficiently ;-)

lib/metadataTypes/Script.js
JoernBerkefeld: Good idea. Let me check the feasibility when I get back to my laptop (travelling atm)

lib/util/util.js
JoernBerkefeld: renamed to extractSingleItem()

JoernBerkefeld: removed method

lib/metadataTypes/Script.js
JoernBerkefeld: done

JoernBerkefeld:
final review please @DougMidgley 
should be all good now

lib/metadataTypes/Discovery.js
```diff
@@ -33,7 +33,7 @@ class Discovery extends MetadataType {
                 v.key = Object.keys(this.definition.endPointMapping)[i];
                 metadataStructure[v.key] = v;
             }
-            await this.saveResults(metadataStructure, retrieveDir, null);
+            await super.saveResults(metadataStructure, retrieveDir, null);
```
DougMidgley, line 36: why did you move to super?

JoernBerkefeld: as discussed, just to signal more clearly that this is actually using the super-version and not a local one. like we've done it in other type-classes.

Issue #263, opened 2022-04-09
[BUG] Import File to All Subs fails


DougMidgley:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when deploying importFile to All Subs, only destinationObjectId is filled, while destinationId is not. This seems to throw a 500 error

### Expected Behavior

deploy to all subs should be supported.

### Version

3.1.3

### Environment

- OS:
- Node:
- npm:

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #264, opened 2022-04-09
#263: Fixes to deploy of Import FIle to all subs

DougMidgley:
# PR details

- [ ] Documentation update
- [X] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [ ] Other, please explain:

additional field being added to create payload of importFile

general validate

## Checklist

- [N/A] README.md updated (if applicable)
- [X] Updated Release Notes

### DIFF
--- a/lib/metadataTypes/definitions/ImportFile.definition.js
+++ b/lib/metadataTypes/definitions/ImportFile.definition.js
@@ -32,7 +32,7 @@ module.exports = {
             isCreateable: true,
             isUpdateable: true,
             retrieving: true,
-            template: false,
+            template: true,
         },
         createdDate: {
             isCreateable: false,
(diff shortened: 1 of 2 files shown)


lib/metadataTypes/definitions/ImportFile.definition.js
```diff
@@ -136,8 +136,8 @@ module.exports = {
         importDefinitionId: {
             isCreateable: false,
             isUpdateable: true,
-            retrieving: true,
-            template: true,
+            retrieving: false,
+            template: false,
```
JoernBerkefeld, line 140: i get why you disable it during retrieve but why during template? if we need it for deploy then surely we'd need it during any templating operation to enable us to redeploy?

JoernBerkefeld (changes requested):

PR #265, opened 2022-04-12
Chore: #180 check blueprint files for changes before trying to override them during upgrade

### LINKED ISSUE #180: [TASK] check blueprint files for changes before trying to override them during `upgrade`
right now it just always wants to override without any checks. That takes longer for the user because they have to confirm it even if there was no change.

goal would be to limit those questions to moments when we would actually change something - and even then potentially have the previous versions available for comparison to auto-update in case it's the expected diff between the existing project version and whats coming with the new update?

JoernBerkefeld:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [x] Other, please explain:
enhanced upgrade process

- closes #180 

...

## Checklist


JoernBerkefeld:
i think this will make it easier with #166 because at least we are now defaulting to override instead of to keeping existing files

DougMidgley (approved):

PR #268, opened 2022-04-13
BREAKING CHANGE #252: retrieve with new auth

### LINKED ISSUE #252: [TASK] Move Auth into dedicated module
Moving auth logic into a dedicated module allows us to better control all aspects of SDK execution, and scopes the relevant code so that it is not mixed with general util logic.

DougMidgley:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [ ] Other, please explain:

moved auth logic into a common lib and aligned with oauth2 naming conventions
updated test classes (nice surprise, we now make one less api call :) )
Also found bug with version force file update when there is no existing file.
closes #252 

Please do an init and run tests yourself also (can be done instead as part of releaes). Still need to update readme, but want review passing first

## Checklist

- [ ] CHANGELOG.md updated


lib/util/auth.js
```diff
@@ -0,0 +1,116 @@
+/**
+ * @typedef {require('../metadataTypes/MetadataType.js')} MetadataType
+ */
+
+/**
+ * @typedef {Object} BuObject
+ * @property {String} credential instance identifier
+ * @property {Integer} mid ID of Business Unit to authenticate with
+ * @property {Integer} eid Parent ID of Business Unit to authenticate with
+ * @property {String} businessUnit name of Business Unit to authenticate with
```
JoernBerkefeld, line 10: simple types (string, integer, boolean) should be lower-case
complex types (Object, Array) should be upper-case

mixed stuff like this: "string[]"

lib/util/init.js
```diff
@@ -16,39 +16,39 @@ const Init = {
...
      * @param {Object} [skipInteraction] signals what to insert automatically for things usually asked via wizard
-     * @param {String} skipInteraction.clientId client id of installed package
-     * @param {String} skipInteraction.clientSecret client secret of installed package
-     * @param {String} skipInteraction.tenant tenant of installed package
-     * @param {String} skipInteraction.eid MID of the Parent Business Unit
-     * @param {String} skipInteraction.credentialsName how you would like the credential to be named
+     * @param {String} skipInteraction.client_id client id of installed package
+     * @param {String} skipInteraction.client_secret client secret of installed package
+     * @param {String} skipInteraction.auth_rul tenant of installed package
+     * @param {String} skipInteraction.account_url MID of the Parent Business Unit
```
JoernBerkefeld, line 24: why rename?

lib/util/util.js
```diff
@@ -193,32 +186,32 @@ const Util = {
...
                                 ) {
-                                    errorMsgs.push(`invalid ClientSecret on ${cred}`);
+                                    errorMsgs.push(`invalid Client Secret on ${cred}`);
                                     solutionSet.add(`Run 'mcdev init ${cred}'`);
                                 }
                                 if (
-                                    !properties.credentials[cred].tenant ||
-                                    properties.credentials[cred].tenant === '--- update me ---'
+                                    !properties.credentials[cred].auth_url ||
+                                    properties.credentials[cred].auth_url === '--- update me ---'
```
JoernBerkefeld, line 205: why the full auth url instead of tenant subdomain? are you planning to save that also going forward?

DougMidgley: auth url is easier to maintain because its copy paste. less parsing :) we dont have a need to have the tenant apart

lib/util/init.js
DougMidgley: because we should align values, where possible, with the ones being used. otherwise we mix up when clientId and when client_id. This standardizes the values. I would note this is a breaking change ;)

lib/util/init.config.js
```diff
@@ -157,10 +157,10 @@ const Init = {
     /**
      * handles creation/update of all config file from the boilerplate
      *
-     * @param {string} versionBeforeUpgrade 'x.y.z'
+     * @param {string} properties config file's json
```
JoernBerkefeld, line 160: this wouldnt work. we overwrite that object before we hit this method and hence would pass in the new version. please revert entire commit

lib/util/auth.js
```diff
@@ -42,10 +42,12 @@ module.exports = {
...
                     Util.logger.warn('Missing scope: ' + missingAccess.join(', '));
                 }
-                const existingAuth = await File.readJson(Util.authFileName);
+                const existingAuth = (await File.pathExists(Util.authFileName))
+                    ? await File.readJson(Util.authFileName)
+                    : {};
                 existingAuth[credential] = authObject;
 
-                File.writeJSON(Util.authFileName, existingAuth);
+                File.writeJSON(Util.authFileName, existingAuth, { spaces: '\t' });
```
JoernBerkefeld, line 50: what for?

lib/util/cli.js
```diff
@@ -393,10 +393,6 @@ const Cli = {
         responses.client_id = responses.client_id.trim();
         responses.client_secret = responses.client_secret.trim();
         responses.auth_url = responses.auth_url.trim();
-        responses.account_id = responses.account_id.trim();
```
JoernBerkefeld: first rename, the fully removing EID?
why? so far that acted as central identifier and second thing we test for to validate credentials are a match between auth and config file

lib/util/file.js
```diff
@@ -504,18 +504,10 @@ const File = {
...
-            auth.credentials[cred] = {};
-            // copy eid as well to make sure we can test for equality when merging the files during runtime
-            auth.credentials[cred].eid = config.credentials[cred].eid;
-        }
         // we want to save to save the full version here to allow us to upgrade configs properly in the future
-        config.version = packageJson.version;
+        properties.version = packageJson.version;
 
-        await this.writeJSONToFile('', Util.configFileName.split('.json')[0], config);
-        await this.writeJSONToFile('', Util.authFileName.split('.json')[0], auth);
```
JoernBerkefeld, line 528: looks to me like you removed storing credentials in a file completely after all? thought we agreed not to push forward with that in this release?

JoernBerkefeld (changes requested):

lib/util/file.js
DougMidgley: It does still save just via the auth handler, to ensure that this logic is handled discretely within that class (auth is managed in auth class). Happy to discuss moving it out and into this, but it also means continued overloading of the file class.

lib/util/cli.js
DougMidgley: its not removing it, it just stops trimming because it is a number, so trim doesnt work :)

lib/util/auth.js
DougMidgley: cf you other comment about removing the file. this is where we save it, and this basically just pretty prints, otherwise its minified ;)

lib/util/init.config.js
DougMidgley: ok, but with my tests it worked exactly as expected, so lets run through one together. alternatively, i can revert and open a bug for fixing, but it makes it impossible for me to test the new init with auth class unless it is resolved

lib/util/init.js
```diff
@@ -102,19 +102,19 @@ const Init = {
                 let responses;
                 if (skipInteraction) {
                     if (
-                        skipInteraction.clientId &&
-                        skipInteraction.clientSecret &&
-                        skipInteraction.tenant &&
-                        skipInteraction.eid &&
-                        skipInteraction.credentialsName
+                        skipInteraction.client_id &&
```
DougMidgley, line 105: add upgrade script

lib/util/cli.js
```diff
@@ -17,11 +18,11 @@ const Cli = {
      * loads default config and adds first credential
      *
      * @param {object} [skipInteraction] signals what to insert automatically for things usually asked via wizard
-     * @param {string} skipInteraction.clientId client id of installed package
-     * @param {string} skipInteraction.clientSecret client secret of installed package
-     * @param {string} skipInteraction.tenant tenant of installed package
+     * @param {string} skipInteraction.client_id client id of installed package
```
DougMidgley, line 21: move to type

lib/util/cli.js
```diff
@@ -300,23 +300,23 @@ const Cli = {
...
+        const status = await BuHelper.refreshBUProperties(properties, inputData.credentialName);
+        return status ? inputData.credentialName : status;
     },
 
     /**
      * helper for addExtraCredential()
      *
      * @param {object} properties from config file
-     * @param {string} [credName] name of credential that needs updating
+     * @param {string} credName name of credential that needs updating
```
DougMidgley, line 311: revert

lib/util/cli.js
```diff
@@ -377,8 +377,8 @@ const Cli = {
                 },
             },
             {
-                type: 'input',
-                name: 'eid',
+                type: 'number',
+                name: 'account_id',
                 message: 'MID of Parent Business Unit',
                 validate: (value) => {
                     if (!Util.isNumeric(value)) {
```
DougMidgley, line 384: remove isNumeric

lib/util/init.config.js
DougMidgley: will update to use version directly due to upgrade scenarios

lib/util/init.js
```diff
@@ -267,10 +267,10 @@ const Init = {
         if (properties && properties.credentials) {
             missingCredentials = Object.keys(properties.credentials).filter(
                 (cred) =>
-                    !properties.credentials[cred].clientId ||
-                    !properties.credentials[cred].clientSecret ||
-                    !properties.credentials[cred].tenant ||
-                    !properties.credentials[cred].eid
+                    !properties.credentials[cred].client_dd ||
```
DougMidgley, line 270: spelling error

lib/util/util.js
```diff
@@ -181,32 +180,32 @@ const Util = {
                                     solutionSet.add('Apply manual fix in your config.');
                                 }
                                 if (
-                                    !properties.credentials[cred].clientId ||
-                                    properties.credentials[cred].clientId === '--- update me ---'
+                                    !properties.credentials[cred].client_id ||
```
DougMidgley, line 183: check how this is managed without credential key

lib/util/init.js
DougMidgley: check proposed upgrade script :)

JoernBerkefeld (approved):

JoernBerkefeld (approved):

PR #274, opened 2022-04-19
Task/172 option to force update of specific config files if we made crucial changes

### LINKED ISSUE #172: Chore: Option to force update of specific config files if we made crucial changes
In case we need to make adjustments to one of the project config files there must be a way to force its upgrade, or at least switch the default to YES for that particular file.

background:
in pr #167 we want to update the SQL formatter but that will require changes to the .vscode/settings.json and to .prettierrc.
Our default is to not override these files to avoid overriding user changes.

_Originally posted by @DougMidgley in https://github.com/Accenture/sfmc-devtools/issues/167#issuecomment-1011468395_

JoernBerkefeld:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [ ] Other, please explain:

closes #172 

image

yes, please check if this works nicely with issue #166 / PR #167

## Checklist


lib/util/init.js
```diff
@@ -239,7 +240,7 @@ const Init = {
         }
 
         // create files before installing dependencies to ensure .gitignore is properly set up
-        status = await InitConfig.createIdeConfigFiles();
+        status = await InitConfig.createIdeConfigFiles(versionBeforeUpgrade);
```
DougMidgley, line 243: if only referenced once, then move versionBeforeUpgrade -> properties.version

lib/util/init.config.js
```diff
@@ -239,13 +251,49 @@ const Init = {
...
+        // !newest entry on top, sorted by version number!
+        let forceIdeConfigUpdate;
+        if (File.existsSync(Util.configFileName)) {
+            try {
+                forceIdeConfigUpdate = File.readJsonSync(
+                    path.resolve(__dirname, Util.boilerplateDirectory, 'forcedUpdates.json')
+                );
+            } catch (ex) {
+                Util.logger.error(`${ex.code}: ${ex.message}`);
+                return [];
```
DougMidgley, line 270: is there a scenario where this wouldnt exist? if not, then lets throw a proper error and fail.

lib/util/init.config.js
```diff
@@ -239,13 +251,49 @@ const Init = {
...
+            }
+        }
+        // return all if no project version was found or only changes from "newer" versions otherwise
+        const relevantForcedUpdates = [];
+        for (let i = 0; i < forceIdeConfigUpdate.length; i++) {
```
DougMidgley, line 275: any reason for not using for of?

lib/util/init.config.js
```diff
@@ -179,7 +184,11 @@ const Init = {
         } else {
             const fileContent = File.readFileSync(gitignoreFileName, 'utf8');
             creationLog.push(
-                await this._createIdeConfigFile(['.' + path.sep, '', '.gitignore'], fileContent)
+                await this._createIdeConfigFile(
+                    ['.' + path.sep, '', '.gitignore'],
```
DougMidgley, line 188: please use path.join or path.resolve here

lib/util/init.config.js
```diff
@@ -198,7 +207,10 @@ const Init = {
                     const ext = '.' + fileArr.pop();
                     // awaiting the result here due to interactive optional overwrite
                     creationLog.push(
-                        await this._createIdeConfigFile([subdir + path.sep, fileArr.join('.'), ext])
+                        await this._createIdeConfigFile(
+                            [subdir + path.sep, fileArr.join('.'), ext],
```
DougMidgley, line 211: potentially here too path.join

DougMidgley (changes requested):

lib/util/init.js
JoernBerkefeld: nope. it gets overwritten in the process of upgrade before we can do our checks. that's why we have to store a backup here

lib/util/init.config.js
JoernBerkefeld: we do that later inside of _createIdeConfigFile - after that function adds some more to the array. running one of the path methods here would make things more complicated and essentially require it to run twice

lib/util/init.config.js
JoernBerkefeld: see other comment on this parameter

lib/util/init.config.js
JoernBerkefeld: none, went old school here but upgraded it now. thx for the reminder :)

lib/util/init.config.js
JoernBerkefeld: that and we kinda need that array down the road:
image

that's why the array is explained in _createIdeConfigFile's jsdoc:
image

JoernBerkefeld:
i added code to ensure overridden config files are backed up. i did once have a developer come crying because somehow they managed to skip version control for some of their configs and lost a whole bunch to the mcdev upgrade command.

PR #284, opened 2022-04-28
Task/283 task replace bluebird lib with p limit

DougMidgley:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [X] Other, please explain: Moved out of bluebird promises to reduce overhead - closes #284 

Removed Bluebird promises(used for concurrency only) and replaced with P-limit

Ensure Tests PR is done first. I tried testing retrieve on Assets and also deploy of query, but would be good if you can double check

## Checklist

- [] CHANGELOG.md updated


lib/metadataTypes/MetadataType.js
```diff
@@ -354,18 +352,24 @@ class MetadataType {
                 Util.metadataLogger('error', this.definition.type, 'upsert', ex, metadataKey);
             }
         }
+        const createLimit = pLimit(10);
```
JoernBerkefeld, line 356: what is this doing?

lib/metadataTypes/MetadataType.js
```diff
@@ -354,18 +352,24 @@ class MetadataType {
...
-                { concurrency: 10 }
+            await Promise.all(
+                metadataToCreate
+                    .filter((r) => r !== undefined && r !== null)
+                    .map((metadataEntry) =>
+                        createLimit(() => this.create(metadataEntry, deployDir))
+                    )
             )
         ).filter((r) => r !== undefined && r !== null);
+        const updateLimit = pLimit(10);
```
JoernBerkefeld, line 366: same as above

package.json
```diff
@@ -45,7 +45,7 @@
         "inquirer": "8.2.2",
         "json-to-table": "4.2.1",
         "mustache": "4.2.0",
-        "p-limit": "4.0.0",
+        "p-limit": "3.1.0",
```
JoernBerkefeld, line 48: why an older version?

JoernBerkefeld (changes requested):

package.json
DougMidgley: same logic, but version 4 uses ESM modules, which was causing issues when running our or CJS context. 
Suggestion would be to hold until a full move to ESM modules OR TS

lib/metadataTypes/MetadataType.js
DougMidgley: This is essentially the rate limiter object, which is applied to each of the iterables. In this case it is limiting it to 10 promises at any given time

lib/metadataTypes/MetadataType.js
DougMidgley: cf. other comment

PR #287, opened 2022-05-02
Task/286 move type definitions into separate files

JoernBerkefeld:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [x] Other, please explain:

moved all type defs into 2 separate files.
this comes with the benefit of being able to use intellisense

had 2 choices here: 

a) disable **jsdoc/no-undefined-types** because eslint somehow cant deal with this the same way vscode's intellisense can.
b) use TYPE = include(types/mcdev.d) everywhere and then prefix all types with "TYPE."
c) move types into util.js assuming it's included everywhere anyways

after fiddling with it I opted for option b to avoid not having checks at all. option c went out the window when I realized that util is in fact not included everywhere yet.

## Checklist

- [ ] CHANGELOG.md updated
- [ ] ran `npm run docs` to update developer docs


lib/util/util.js
```diff
@@ -43,6 +25,7 @@ const Util = {
     parentBuName: '_ParentBU_',
     standardizedSplitChar: '/',
     skipInteraction: false,
+    SDK: SDK,
```
DougMidgley, line 28: double check this

DougMidgley (changes requested):

DougMidgley (approved):

Issue #18, opened 2021-05-05
include buildTemplate in createDeltaPkg instead of retrieveAsTemplate / CSCLSROZ-555


SabrinaEspino:
the current approach does not actually use the files committed to git but retrieves again from the server -that might break things given that DEV might have already changed by the time this runs.

therefore, buildTemplate must be added which does what retrieveAsTemplate does, only completely local, without a retrieve

JoernBerkefeld:
was duplicated by #19 - closed the other one to work in this one

PR #299, opened 2022-05-06
Feature/19 add buildtemplate based on rt command that uses retrieve folder and not server as source

### LINKED ISSUE #19: add `buildTemplate` based on `rt` command that uses retrieve folder and not server as source / CSCLSROZ-230
the current approach does not actually use the files committed to git but retrieves again from the server -that might break things given that DEV might have already changed by the time this runs.

therefore, **buildTemplate** must be added which does what retrieveAsTemplate does, only completely local, without a retrieve

JoernBerkefeld:
# PR details

- [x] Add a CLI option

- closes #19 
  - added buildTemplate command with **support for all types**
  - this blocks #18 (upgrade of `createDeltaPkg`) from being closed

we should discuss deprecating `retrieveAsTemplate` or, at the very least, rewrite `retrieveAsTemplate` to be a mere wrapper around buildTemplate that downloads given metadata

## Checklist


JoernBerkefeld:
other PRs are merged. this one's ready for review now @DougMidgley  :)

lib/Builder.js
```diff
@@ -90,6 +99,55 @@ class Builder {
...
+     * Builds a specific metadata file by name
+     *
+     * @param {string} metadataType metadata type to create a template of
+     * @param {string} key customerkey of metadata to create a template of
+     * @param {TYPE.TemplateMap} templateVariables variables to be replaced in the metadata
+     * @returns {Promise.<TYPE.MultiMetadataTypeList>} Promise
+     */
+    async buildTemplate(metadataType, key, templateVariables) {
+        let keyArr;
+        if (key.includes(',')) {
```
DougMidgley, line 112: ternary would be better suited here :) (prefer const)

lib/Builder.js
```diff
@@ -90,6 +99,55 @@ class Builder {
...
+
+                    return MetadataTypeInfo[type].buildTemplate(
+                        this.retrieveDir,
+                        this.templateDir,
+                        key,
+                        templateVariables
+                    );
+                })
+            );
+            // TODO confirm if should replace this.metadata
```
DougMidgley, line 139: please do this then :)

lib/Builder.js
```diff
@@ -90,6 +99,55 @@ class Builder {
...
+            if (result) {
+                this.metadata[result[0].type] = [];
+                result.forEach((element) => {
```
DougMidgley, line 142: forEach + push = map :)

lib/cli.js
```diff
@@ -185,6 +185,36 @@ yargs
...
+                .positional('BU', {
+                    type: 'string',
+                    describe:
+                        'the business unit to deploy to (in format "credential name/BU name")',
+                })
+                .positional('TYPE', {
+                    type: 'string',
+                    describe: 'metadata type',
+                })
+                .positional('KEY', {
```
DougMidgley, line 203: if can accept multiple then KEYS would be better i guess :)

lib/cli.js
```diff
@@ -185,6 +185,36 @@ yargs
...
+                })
+                .positional('MARKET', {
+                    type: 'string',
+                    describe: 'market used for reverse building template',
+                });
+        },
+        handler: (argv) => {
+            Mcdev.setSkipInteraction(!!argv.skipInteraction);
+            Util.setLoggingLevel(argv);
+            Mcdev.buildTemplate(argv.BU, argv.TYPE, argv.KEY, argv.MARKET);
```
DougMidgley, line 215: i think turning into array here would be better :)

lib/metadataTypes/Asset.js
```diff
@@ -476,26 +476,106 @@ class Asset extends MetadataType {
...
+     * @returns {Promise.<void>} void
+     */
+    static buildTemplateForExtracts(
+        templateDir,
+        targetDir,
+        metadata,
+        templateVariables,
+        templateName
+    ) {
+        return this._buildXForExtracts(
```
DougMidgley, line 548: can we have a chat on this logic so i get it?

lib/metadataTypes/Asset.js
```diff
@@ -515,6 +607,7 @@ class Asset extends MetadataType {
                 );
             }
         }
+        console.log('fileList', fileList);
```
DougMidgley, line 610: nope

lib/metadataTypes/Query.js
```diff
@@ -116,45 +116,129 @@ class Query extends MetadataType {
...
+     * helper for buildDefinitionForExtracts
+     * searches extracted SQL file for template variables and applies the market values
+     *
+     * @param {string} code code from extracted code
+     * @param {TYPE.TemplateMap} templateVariables variables to be replaced in the metadata
+     * @returns {string} code with markets applied
+     */
+    static _applyMarketValues(code, templateVariables) {
+        // fix bad formatting applied by SQL Formatter Plus
+        code = code
```
DougMidgley, line 129: this already exists somewhere else i think, if the same, then please merge into util method

test/dataExtension.test.js
```diff
@@ -3,99 +3,150 @@ const cache = require('../lib/util/cache');
...
-            await testUtils.getActualFile('childBU_dataextension_test', 'dataExtension'),
-            'returned metadata was not equal expected'
-        );
-        assert.equal(
-            Object.values(testUtils.getAPIHistory()).flat().length,
-            6,
-            'Unexpected number of requests made'
-        );
-        return;
+    describe('Retrieve ================', () => {
```
DougMidgley, line 14: if this is being restructured here, please also do for query OR leave as is

DougMidgley (changes requested):
Generally looks ok, but I need a walk through to question the approach in some cases

lib/metadataTypes/Asset.js
JoernBerkefeld: Before the talk: we already had a buildDefinitionForExtract. That included the logic which files to look for. The only diff to what we needed for buildTEMPLATE was running mustache vs. replacing dev value with market variable names

lib/cli.js
JoernBerkefeld: Single key is default case, multiple keys the advanced case. I'd rather go with the simple default case when naming variables

lib/Builder.js
JoernBerkefeld: Todo: remove comment

JoernBerkefeld:
**TODO:**
update README.md to state that templating is now supported for ALL types (via bt, but not via rt)

lib/metadataTypes/Query.js
```diff
@@ -116,45 +116,117 @@ class Query extends MetadataType {
...
-    static async buildDefinitionForExtracts(
+    static buildDefinitionForExtracts(
         templateDir,
         targetDir,
         metadata,
-        variables,
+        templateVariables,
         templateName
+    ) {
+        return this._buildXForExtracts(
```
DougMidgley, line 160: why not call this directly? these helper methods dont needs to be that helpfu? (will continue reading other methods maybe it does make sense)

lib/metadataTypes/Query.js
```diff
@@ -116,45 +116,117 @@ class Query extends MetadataType {
...
+     * handles extracted code if any are found for complex types
+     *
+     * @param {string} templateDir Directory where metadata templates are stored
+     * @param {string|string[]} targetDir (List of) Directory where built definitions will be saved
+     * @param {TYPE.QueryItem} metadata main JSON file that was read from file system
+     * @param {TYPE.TemplateMap} templateVariables variables to be replaced in the metadata
+     * @param {string} templateName name of the template to be built
+     * @returns {Promise.<void>} void
+     */
+    static buildTemplateForExtracts(
```
DougMidgley, line 180: suggestions for new name for clarity

- buildTemplateForNested
- extractNestedForTemplate

As it stands it sounds a bit like a data extract, and having then buildXForExtracts makes it worse. 
buildXForExtracts, could be then
- buildForNested
- buildNestedMetadata
for example

lib/metadataTypes/Script.js
```diff
@@ -141,32 +140,92 @@ class Script extends MetadataType {
...
+        templateVariables,
+        templateName,
+        mode
     ) {
         // get SSJS from filesystem
         let code = await this._mergeCode(metadata, templateDir, templateName);
-        // replace template variables with their values
         try {
-            code = Mustache.render(code, variables);
+            if (mode === 'definition') {
```
DougMidgley, line 221: I would probably make this a ternary operator given we only have two scenarios? otherwise, i would expect an else block with an error

DougMidgley (changes requested):

lib/metadataTypes/Script.js
JoernBerkefeld: I find it easier to understand the logic given that this isbthe one and only place where we do sth different for buildTemplate vs buildDefinition. The fact that it is an internal helper and even jsdoc defined proper values leads me to think that an else-block is wasting lines here?

lib/metadataTypes/Query.js
JoernBerkefeld: We could but think about how you struggled with understanding the logic here even with that in place

JoernBerkefeld:
@DougMidgley i made the requested changes aside from the 2 comments above. please have a look :)

JoernBerkefeld:
merged without repeated review assuming all requested changes were implemented in good conscience

PR #301, opened 2022-05-06
Feature: #18 include buildtemplate in createdeltapkg + #300 no more key-name swapping in rt

### LINKED ISSUE #18: include buildTemplate in createDeltaPkg instead of retrieveAsTemplate / CSCLSROZ-555

### LINKED ISSUE #300: [BREAKING CHANGE] Disable key/name override during retrieveAsTemplate
this is malfunctioning if key & name are not equal and is overall causing confusion.

JoernBerkefeld:
# PR details

- [ ] Documentation update
- [ ] Bug fix
- [ ] New metadata support
- [ ] Enhanced metadata
- [ ] Add a CLI option
- [ ] Other, please explain:

- closes #300 (breaking change)
- closes #18 
-

**please merge PR #299 first.**

## Checklist

### DIFF
--- a/lib/Builder.js
+++ b/lib/Builder.js
@@ -139,6 +139,7 @@ class Builder {
     async buildTemplate(metadataType, keyArr, templateVariables) {
         const type = metadataType;
         try {
+            /** @type {TYPE.MetadataTypeItemObj[]} */
             const result = await Promise.all(
                 keyArr.map((key) => {
                     // with npx and powershell spaces are not parsed correctly as part of a string
@@ -148,6 +149,7 @@ class Builder {
                     MetadataTypeInfo[type].client = this.client;
                     MetadataTypeInfo[type].properties = this.properties;
 
+                    /** @type {TYPE.MetadataTypeItemObj} */
                     return MetadataTypeInfo[type].buildTemplate(
                         this.retrieveDir,
                         this.templateDir,
--- a/lib/Builder.js
+++ b/lib/Builder.js
@@ -118,7 +118,7 @@ class Builder {
         const buObject = await Cli.getCredentialObject(properties, businessUnit);
         if (buObject !== null) {
             const builder = new Builder(properties, buObject);
-            if (this._checkMarket(market)) {
+            if (Util.checkMarket(market, properties)) {
                 return builder.buildTemplate(selectedType, keyArr, properties.markets[market]);
             }
         }
@@ -184,14 +184,8 @@ class Builder {
         const buObject = await Cli.getCredentialObject(properties, businessUnit);
         if (buObject !== null) {
             const builder = new Builder(properties, buObject);
-            if (market === '*') {
-                for (const oneMarket in properties.markets) {
-                    builder.buildDefinition(selectedType, name, properties.markets[oneMarket]);
-                }
-            } else {
-                if (this._checkMarket(market)) {
-                    builder.buildDefinition(selectedType, name, properties.markets[market]);
-                }
+            if (Util.checkMarket(market, properties)) {
+                builder.buildDefinition(selectedType, name, properties.markets[market]);
             }
         }
     }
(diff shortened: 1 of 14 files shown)


JoernBerkefeld:
@DougMidgley this one is ready for review now

lib/Builder.js
```diff
@@ -178,6 +294,28 @@ class Builder {
             }
         }
     }
+    /**
+     * check if a market name exists in current mcdev config
+     *
+     * @param {string} market market localizations
+     * @returns {boolean} found market or not
+     */
+    static _checkMarket(market) {
```
DougMidgley, line 303: this feels a lot like it should be in util not in builder?

lib/Builder.js
```diff
@@ -178,6 +294,28 @@ class Builder {
...
+        const properties = File.loadConfigFile();
+        if (properties.markets[market]) {
+            return true;
+        } else {
+            Util.logger.error(`Could not find the market '${market}' in your configuration file.`);
+            const marketArr = [];
+            for (const oneMarket in properties.markets) {
```
DougMidgley, line 310: const martketArr = Object.values(properties.markets) would be preferred

lib/util/devops.js
```diff
@@ -186,7 +191,7 @@ const DevOps = {
         }
         // Write into delta.json to serve as documentation
         File.writeJSONToFile(properties.directories.deltaPackage, 'delta_package', delta);
-        this.document(properties.directories.deltaPackage, delta);
+        DevOps.document(properties.directories.deltaPackage, delta);
```
DougMidgley, line 194: any reason to change this?

lib/util/devops.js
```diff
@@ -237,6 +243,178 @@ const DevOps = {
...
+                            name: 'isPurgeDeployFolder',
+                            message:
+                                'Do you want to empty the deploy folder (ensures no files from previous deployments remain)?',
+                            default: true,
+                        },
+                    ]);
+                }
+                if (skipInteraction || responses.isPurgeDeployFolders) {
+                    // Clear output folder structure for selected sub-type
+                    File.removeSync(File.normalizePath([properties.directories.deploy]));
```
DougMidgley, line 389: prefer async

lib/util/devops.js
```diff
@@ -237,6 +243,178 @@ const DevOps = {
...
         return delta;
     },
+    /**
+     * wrapper around DevOps.getDeltaList, Builder.buildTemplate and M
+     *
+     * @param {object} properties project config file
+     * @param {string} range git commit range
+     * @param {TYPE.SkipInteraction} [skipInteraction] allows to skip interactive wizard
+     */
+    async buildDeltaDefinitions(properties, range, skipInteraction) {
```
DougMidgley, line 253: part of me feels like this may benefit from being broken down into a couple of diff methods.

DougMidgley (changes requested):

lib/util/devops.js
JoernBerkefeld: nope, not really. consistency maybe

lib/util/devops.js
JoernBerkefeld: i agree. not really in scope here though but let me create a ticket.

JoernBerkefeld: created #318

lib/Builder.js
DougMidgley: yes, but in the utils file, you could avoid getting loadConfigFile, and just pass the properties in. I looked at all references and all have properties object available before using this method. I can imagine it being useful for other items in future elsewhere :/ not about to fall on my sword about it given its only open item - we could reopen in a new point about removing this cyclical dependency item

DougMidgley (changes requested):

lib/Builder.js
JoernBerkefeld: done

DougMidgley (approved):

PR #324, opened 2022-05-19
Task/294 task run checkproperties on config load

DougMidgley:
# PR details

## What is the purpose of this pull request?

- [X] Other, please explain: Internal

rework loading parameters logic into a new class to help reduce circular dependencies and remove bloat in common classes

initial stab - feedback requested more generally about approach

## Checklist

- [ ] ESLint & Prettier are not outputting errors or warnings
- [ ] CHANGELOG.md updated
- [ ] ran `npm run docs` to update developer docs


JoernBerkefeld:
Wrong target? ;)

lib/util/cli.js
```diff
@@ -85,7 +86,7 @@ const Cli = {
      */
     async getCredentialObject(properties, target, isCredentialOnly, allowAll) {
         try {
-            if (!(await Util.checkProperties(properties))) {
+            if (!(await Config.checkProperties(properties))) {
```
JoernBerkefeld, line 89: does this logic still work, given that you run checkProperties inside of getProperties already?

DougMidgley: I assume you mean, is it needed? I would say actually no, so we can throw errors earlier on getProperties?

lib/util/cli.js
```diff
@@ -38,7 +39,7 @@ const Cli = {
      * @returns {Promise.<void>} -
      */
     async addExtraCredential(properties, skipInteraction) {
-        if (!(await Util.checkProperties(properties))) {
+        if (!(await Config.checkProperties(properties))) {
```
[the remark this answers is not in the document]
DougMidgley: ditto other comment

lib/util/cli.js
```diff
@@ -7,6 +7,7 @@ const inquirer = require('inquirer');
 const MetadataDefinitions = require('./../MetadataTypeDefinitions');
 const Util = require('./util');
 const auth = require('./auth');
+const Config = require('./config');
```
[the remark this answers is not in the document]
DougMidgley: Good point, reset to lower

lib/index.js
```diff
@@ -43,10 +43,7 @@ class Mcdev {
      * @returns {Promise.<TYPE.DeltaPkgItem[]>} list of changed items
      */
     static async createDeltaPkg(argv) {
-        properties = properties || File.loadConfigFile();
-        if (!(await Util.checkProperties(properties))) {
-            return null;
-        }
```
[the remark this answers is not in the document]
DougMidgley: would it work to use throw error in checkProperties, can also add missing fields as an error property for that use that. I think that would clean up a couple of things

lib/util/config.js
```diff
@@ -0,0 +1,289 @@
...
+                        );
+                    }
+                }
+            } else if (!silent) {
+                Util.logger.warn(
+                    `${Util.authFileName} not found. Please run 'mcdev init' to provide the missing credential details.`
+                );
+            }
+        }
+        Config.checkProperties(Config.properties);
```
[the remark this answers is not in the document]
DougMidgley: added await - if we move to throw error should be ok in this context

lib/index.js
```diff
@@ -351,7 +352,7 @@ class Mcdev {
      */
     static async deleteByKey(businessUnit, type, customerKey) {
         Util.logger.info('mcdev:: delete');
-        properties = properties || File.loadConfigFile();
+        const properties = await config.getProperties();
```
JoernBerkefeld, line 355: depending on how you want to see it either this or cli.js is highest level of execution. neither are handling potential errors returned by your new method at this point. hard no!

lib/util/cli.js
```diff
@@ -37,24 +38,22 @@ const Cli = {
...
-                    Util.logger.info(` - ${cred}`);
-                }
-            }
-            Util.logger.info('\nPlease enter your new credentials');
-            if (skipInteraction && properties.credentials[skipInteraction.credentialName]) {
-                Util.logger.error(
-                    `Credential '${skipInteraction.credentialName}' already existing. If you tried updating please provide run 'mcdev init ${skipInteraction.credentialName}'`
-                );
+        // check existing properties before adding additional // todo CONFIRM if needed
+        await config.checkProperties(properties);
```
JoernBerkefeld, line 42: is an error here handled higher up in the stack? does running this even make sense, considering you are running it inside of getProperties()?

lib/Builder.js
```diff
@@ -103,7 +104,7 @@ saved
      */
     static async buildTemplate(businessUnit, selectedType, keyArr, market) {
         Util.logger.info('mcdev:: Build Definition from Template');
-        const properties = File.loadConfigFile();
+        const properties = await config.getProperties();
```
JoernBerkefeld, line 107: not handled

lib/Builder.js
```diff
@@ -170,7 +171,7 @@ saved
      * @returns {Promise.<TYPE.MultiMetadataTypeList>} -
      */
     static async buildDefinition(businessUnit, selectedType, name, market) {
-        const properties = File.loadConfigFile();
+        const properties = await config.getProperties();
```
JoernBerkefeld, line 174: not handled

lib/Deployer.js
```diff
@@ -138,7 +140,7 @@ class Deployer {
     static async _deployBU(cred, bu, typeArr, keyArr) {
         const buPath = `${cred}/${bu}`;
         Util.logger.info(`::Deploying ${buPath}`);
-        const properties = File.loadConfigFile();
+        const properties = await config.getProperties();
```
JoernBerkefeld, line 143: not handled

JoernBerkefeld (changes requested):
lots still to do here. I'd ask that we please follow through with my PRs first and once this reaches production readiness I'd happily adjust to the new reality. 

our tool has to work properly and have reasonably nice error output for users in a CLI. Therefore, we cannot just throw errors without handling them. Prefixing our errors with "Retrieve failed: ...", "Document failed: ...." is also not very user friendly. For one, the user just executed that command. whatelse should have failed? Therefore, whatever error message comes needs to be printable as is - and I've gone through great lengths to make that happen over the past months.

The few times you added try-catch blocks seem like a step back in that regard. For the many other lines on which you had to replace getting the config, error handling is just missing and would result in ugly stack traces being presented to our already not-so-CLI-friendly user base.

**please note the 11+ hidden conversations**

lib/index.js
```diff
@@ -90,16 +84,19 @@ class Mcdev {
      */
     static async upgrade(skipInteraction) {
         Mcdev.setSkipInteraction(skipInteraction);
-        properties = properties || File.loadConfigFile();
+        const properties = await config.getProperties();
         if (!properties) {
             Util.logger.error('No config found. Please run mcdev init');
             return false;
         }
```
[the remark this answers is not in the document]
DougMidgley: removing as already handled

lib/index.js
```diff
@@ -72,10 +69,7 @@ class Mcdev {
      * @returns {Promise} .
      */
     static async selectTypes() {
-        properties = properties || File.loadConfigFile();
-        if (!(await Util.checkProperties(properties))) {
-            return null;
-        }
+        const properties = await config.getProperties();
```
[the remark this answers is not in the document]
DougMidgley: moved to cli as discussed

lib/util/cli.js
JoernBerkefeld: push

lib/util/config.js
```diff
@@ -0,0 +1,274 @@
...
+            errorMsgs.push(
+                `Your project's config version ${properties.version} is lower than your Accenture SFMC DevTools version ${packageJsonMcdev.version}`
+            );
+            solutionSet.add(`Run 'mcdev upgrade' to ensure optimal performance`);
+            missingFields.push('version');
+        }
+        if (silent) {
+            const err = new Error('Missing Fields');
+            err.missingFields = missingFields;
+            throw err;
```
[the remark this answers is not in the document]
JoernBerkefeld: this feels like a "lets sit at the same desk and talk" topic

lib/util/config.js
```diff
@@ -0,0 +1,274 @@
...
+                        name: 'runUpgradeNow',
+                        message: `Do you want to run 'mcdev upgrade' now?`,
+                        default: true,
+                    },
+                ]);
+                if (responses.runUpgradeNow) {
+                    // use _execSync here to avoid a circular dependency
+                    this.execSync('mcdev', ['upgrade']);
+                }
+                throw new Error(errorMsgOutput);
```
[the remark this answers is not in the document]
JoernBerkefeld: after we've run mcdev we should end execution. we COULD continue automatically but ensuring that this runs smoothly then might just not be worth the effort

lib/index.js
```diff
@@ -313,32 +306,29 @@ class Mcdev {
...
+                parentBUOnlyTypes.includes(type) ? Util.parentBuName : null
+            );
+            if (buObject !== null) {
+                MetadataTypeInfo[type].properties = properties;
+                MetadataTypeInfo[type].document(buObject);
             }
         } catch (ex) {
-            Util.logger.error(`Document failed: ${ex.message}`);
+            Util.logger.error('mcdev.document ' + ex.message);
+            Util.logger.debug(ex.stack);
```
JoernBerkefeld, line 329: please use this instead
https://github.com/Accenture/sfmc-devtools/blob/bc01163fe4a30c16a00f660497c585aaed13bf96/lib/util/util.js#L474

lib/index.js
```diff
@@ -313,32 +306,29 @@ class Mcdev {
      */
     static async document(businessUnit, type) {
         Util.logger.info('mcdev:: Document');
+
+        const properties = await config.getProperties();
+        if (type && !MetadataTypeInfo[type]) {
+            Util.logger.error(`:: '${type}' is not a valid metadata type`);
+            return;
+        }
```
JoernBerkefeld, line 315: should be done before getting the config to increase speed of execution.
also, given that this is highest level execution - should we not switch the error log into a throw error instead?

lib/index.js
```diff
@@ -313,32 +306,29 @@ class Mcdev {
...
+                MetadataTypeInfo[type].properties = properties;
+                MetadataTypeInfo[type].document(buObject);
             }
         } catch (ex) {
-            Util.logger.error(`Document failed: ${ex.message}`);
+            Util.logger.error('mcdev.document ' + ex.message);
+            Util.logger.debug(ex.stack);
+            Util.logger.info(
+                'If the directoy does not exist, you may need to retrieve this BU first.'
+            );
```
JoernBerkefeld, line 332: maybe throw error instead given the new catch in cli.js?

lib/cli.js
```diff
@@ -374,6 +374,10 @@ yargs
     })
     .demandCommand(1, 'Please enter a valid command')
     .strict()
+    .fail((_, err) => {
+        Util.logger.error(err.message);
+        process.exit(1);
```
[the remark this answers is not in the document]
JoernBerkefeld: https://github.com/Accenture/sfmc-devtools/blob/bc01163fe4a30c16a00f660497c585aaed13bf96/lib/util/util.js#L499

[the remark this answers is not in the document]
JoernBerkefeld: also there is a method to wrap around exit code 1:
https://github.com/Accenture/sfmc-devtools/blob/bc01163fe4a30c16a00f660497c585aaed13bf96/lib/util/util.js#L131

lib/util/config.js
```diff
@@ -0,0 +1,275 @@
...
+ */
+const config = {
+    properties: null,
+    /**
+     * loads central properties from config file
+     *
+     * @param {boolean} [silent] omit throwing errors and print messages; assuming not silent if not set
+     * @returns {object} central properties object
+     */
+    async getProperties(silent) {
```
JoernBerkefeld, line 18: note - this is how we define methods in other util classes

lib/util/config.js
```diff
@@ -0,0 +1,275 @@
...
+            }
+        }
+    },
+    /**
+     * defines how the properties.json should look like
+     * used for creating a template and for checking if variables are set
+     *
+     * @returns {object} default properties
+     */
+    getDefaultProperties: async function () {
```
JoernBerkefeld, line 259: and again the other type of notation.
this sets a property to an anonymous method rather than define the method getDefaultProperties as part of the Config object

JoernBerkefeld (changes requested):

lib/util/config.js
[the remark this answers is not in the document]
JoernBerkefeld: went through the code again but cant help but think that wrapping this in an error adds no value here. It does add extra lines of code though and complexity. we can argue about renaming the parameter "silent" to "returnList" or sth like it but that's really the expected outcome here and while it's true that you could do that with try-catch and attaching it to the error, there are probably also 20 other ways but none are better than simply returning the array that the other method is looking for here.

lib/util/config.js
```diff
@@ -12,10 +13,10 @@ const config = {
     /**
      * loads central properties from config file
      *
-     * @param {boolean} [silent] omit throwing errors and print messages; assuming not silent if not set
+     * @param {boolean} [skipChecks] omit throwing errors and print messages
      * @returns {object} central properties object
      */
-    async getProperties(silent) {
+    async getProperties(skipChecks = false) {
```
DougMidgley, line 19: again, i would say withChecks = true here to avoid double negatives. https://blog.devgenius.io/code-smell-51-double-negatives-993b6160f3e1

PR #350, opened 2022-06-22
Task/346 only upsert metadata that was actually changed

### LINKED ISSUE #346: [TASK] only upsert metadata that was actually changed
right now, whatever folder we might need is updated, which does take a second per folder. large deployments are unnecessarily slow due to that.

instead we could check the cache for actual changes, like we do for DE fields

JoernBerkefeld:
# PR details

- closes #346 

so far only supports folders but that can already be a big help. future versions can add more type support for this feature.

## Checklist


lib/metadataTypes/MetadataType.js
```diff
@@ -311,6 +311,59 @@ class MetadataType {
...
+    /**
+     * test if metadata was actually changed or not to potentially skip it during deployment
+     *
+     * @param {TYPE.MetadataTypeItem} cachedVersion cached version from the server
+     * @param {TYPE.MetadataTypeItem} metadata item to upload
+     * @param {string} [fieldName] optional field name to use for identifying the record in logs
+     * @param {boolean} [silent] optionally suppress logging
+     * @returns {boolean} true on first identified deviation or false if none are found
+     */
+    static hasChanged_generic(cachedVersion, metadata, fieldName, silent) {
```
DougMidgley, line 335: naming convention would be hasChangedGeneric i guess

lib/metadataTypes/MetadataType.js
```diff
@@ -311,6 +311,59 @@ class MetadataType {
...
+        }
+        // we do need the full set in other places and hence need to work with a clone here
+        const clonedMetada = JSON.parse(JSON.stringify(metadata));
+        this.removeNotUpdateableFields(clonedMetada);
+        // iterate over what we want to upload rather than what we cached to avoid false positives
+        for (const prop in clonedMetada) {
+            if (this.definition.ignoreFieldsForUpdateCheck?.includes(prop)) {
+                continue;
+            }
+            if (Util.isObjectChanged(cachedVersion, metadata, prop)) {
```
DougMidgley, line 347: there is helper libraries which do this. including "assert" which we use for testing. i would recommend we reuse that one, rather than build our own

DougMidgley (changes requested):

lib/metadataTypes/MetadataType.js
JoernBerkefeld: if you can point me towards the right methods? i found it to be difficult given our complex cases

lib/Deployer.js
```diff
@@ -20,12 +20,13 @@ class Deployer {
      *
      * @param {TYPE.Mcdevrc} properties General configuration to be used in retrieve
      * @param {TYPE.BuObject} buObject properties for auth
+     * @param {boolean} [fromRetrieve] optionally deploy whats defined via selectedTypesArr + keyArr directly from retrieve folder instead of from deploy folder
      */
-    constructor(properties, buObject) {
```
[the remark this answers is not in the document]
JoernBerkefeld: wrong PR?

lib/metadataTypes/MetadataType.js
DougMidgley: For example https://nodejs.org/api/assert.html#assertdeepequalactual-expected-message or https://nodejs.org/api/assert.html#assertequalactual-expected-message could cover this need.

lib/metadataTypes/MetadataType.js
```diff
@@ -311,6 +311,59 @@ class MetadataType {
...
 
+    /**
+     * test if metadata was actually changed or not to potentially skip it during deployment
+     *
+     * @param {TYPE.MetadataTypeItem} cachedVersion cached version from the server
+     * @param {TYPE.MetadataTypeItem} metadata item to upload
+     * @param {string} [fieldName] optional field name to use for identifying the record in logs
+     * @returns {boolean} true if metadata was changed
+     */
+    static hasChanged(cachedVersion, metadata, fieldName) {
```
DougMidgley, line 322: should we not just use hasChangedGeneric and if it is overriden then the override applies?

JoernBerkefeld: ideally yes. but we have too many cases where whats in cache doesn't necessarily match what we deploy. therefore, i'd rather opt-in per type than opt-out.

lib/util/util.js
```diff
@@ -641,6 +641,71 @@ const Util = {
...
+        for (let i = 0, l = arrayA.length; i < l; i++) {
+            // Check if we have nested arrays
+            if (Array.isArray(arrayA[i]) || Array.isArray(arrayB[i])) {
+                // recurse into the nested arrays
+                if (!this.arrayEquals(arrayA[i], arrayB[i])) {
+                    return false;
+                }
+            } else if (arrayA[i] != arrayB[i]) {
+                // Warning - two different object instances will never be equal: {x:20} != {x:20}
+                return false;
```
DougMidgley, line 704: if we consider automations, objects would be nested in arrays. So this would force it to always be deployed, when actually that may not be needed.  is that the correct behaviour? https://www.npmjs.com/package/deep-equal would still seem to be a better way to actually check diffs

DougMidgley (approved):

Issue #162, opened 2022-01-11
[FEATURE] print time it took to run retrieve / deploy command (including * versions)


JoernBerkefeld:
idea is that whenever relevant methods are executed, the user sees how long it took to do so

JoernBerkefeld:
code is async, means extra complexity to time it without interfering with performance

what to time: methods in lib/index.js
- retrieve
- deploy
- retrieveAsTemplate
- buildDefinition
- buildDefinitionBulk
- createDeltaPkg

JoernBerkefeld:
check if framework **yargs** (see lib/cli.js) already has an option for that:
https://github.com/yargs/yargs

JoernBerkefeld:
final solution only covers 
- retrieve
- deploy
- retrieveAsTemplate

rest is run locally anyways within seconds

PR #389, opened 2022-07-22
#162: print time it takes to run retrieve/deploy command

tuliodnw:
# PR details

- [x] Other, please explain:

added 2 functions in util, one to get the time in ms before function and a second one that calculates and prints the time passed between the first function in ms.

the stopTimer in index.js, is adding a stopTimer before every return the best way?

## Checklist

- [x] ESLint & Prettier are not outputting errors or warnings  //after removing the warnings and saving, the warnings appear again


JoernBerkefeld:
There is always the option to have it in index.js or in cli.js - but let's check this out first the way you built it.

JoernBerkefeld:
not going to use this solution. new PR will be created

Issue #518, opened 2022-11-06
[BUG] check if we are missing any types


JoernBerkefeld:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Our internal list might be missing a few things. email templates was not included until #515 for instance

### Expected Behavior

https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/base-asset-types.html

### Steps To Reproduce

3. Run '...'
4. See error...

### Version

4.1.9

### Environment

- OS:
- Node:
- npm:

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #524, opened 2022-11-07
#518 changing Asset.definition.js

DannyCastroVentura:
# PR details

- [X] New metadata support

- Added the element's name on every extendedSubTypes, as it was on template
- For this # 1 to be completed I had to add the file, custom and default elements to the extendedSubTypes
- Added items on the typeMapping
- Reordered some elements ("template: 4" in the typeMapping, and swapped order with "message" and "template")
## Checklist

- [X] ran `npm run test` to with 0 tests failing

### DIFF
--- a/lib/metadataTypes/definitions/Asset.definition.js
+++ b/lib/metadataTypes/definitions/Asset.definition.js
@@ -461,6 +461,17 @@ module.exports = {
     binarySubtypes: ['archive', 'audio', 'document', 'image', 'rawimage', 'video'],
     extendedSubTypes: {
         asset: ['webpage', 'webtemplate', 'jsonmessage', 'icemailformblock', 'asset'],
+        file: [
+            'image',
+            'rawimage',
+            'video',
+            'document',
+            'audio',
+            'archive',
+            'code',
+            'textfile',
+            'file',
+        ],
         image: [
             'ai',
             'psd',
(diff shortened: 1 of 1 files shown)


lib/metadataTypes/definitions/Asset.definition.js
```diff
@@ -460,7 +460,18 @@ module.exports = {
...
+            'image',
+            'rawimage',
+            'video',
+            'document',
+            'audio',
+            'archive',
+            'code',
+            'textfile',
+            'file',
+        ],
```
JoernBerkefeld, line 474:
```suggestion
```
please remove

lib/metadataTypes/definitions/Asset.definition.js
```diff
@@ -667,9 +685,12 @@ module.exports = {
             'socialfollowblock',
             'buttonblock',
             'layoutblock',
+            'block',
         ],
-        message: ['templatebasedemail', 'htmlemail', 'textonlyemail'],
         template: ['defaulttemplate', 'template'],
+        message: ['templatebasedemail', 'htmlemail', 'textonlyemail', 'message'],
+        custom: ['custom'],
+        default: ['default'],
```
JoernBerkefeld, line 693:
```suggestion
```
i think we will need to remove this as well

JoernBerkefeld (changes requested):
with your changes this was my retrieve result: **430 assets downloaded**
image


and here without your changes: **430 assets downloaded**
image

I suppose we can summarize it didn't break anything and didn't make anything worse - given that my test server did not include every type I think this is a good result.

lets please apply the suggested changes and then I think we are good to merge this.
why remove some? because it leads to duplication otherwise. the type "file"'s subtypes are already covered elsewhere and "custom" and "default" seem to be placeholders

PR #549, opened 2022-11-10
#542 fixing problem with not valid customer key on assets when deploying emailSendDefinition

### LINKED ISSUE #542: [BUG] emailSendDefinition deploy breaks due to missing email asset / list

### Current Behavior

**issue 1**
while default templating would have been applied to the json as part of running buildDefinition, we also add the target MID to asset keys during deploy.
that causes deploy of emailSendDefinitions to fail because when it is processed, it cant find the email-asset in cache with its key before the MID-renaming.

**issue 2**
_additionally, depending on BU settings, if All Subscriber was set up as relevant list, that list's id might be of the all-subs list on the parent bu or the one on the current BU. As of now we cannot tell which it is and hence cannot decide how to deploy it_ 
**note:** issue 2 was solved already generically by updating the caching function for lists

### Expected Behavior

somehow ensure that we can find the email in cache

### Steps To Reproduce

1. create an email in content builder and a email activity in automation studio
2. try to deploy both (or one at a time) to another BU
3. See error...

### Version

4.1.11

### Environment

_No response_

### Participation

### Additional comments

_No response_

DannyCastroVentura:
**tickets:**
- fixes #542 

**dependencies**
- blocked by pr #562
- blocked by pr #578
- blocked by pr #592 
- blocked by pr #595 

fixing problem with not valid customer key on assets when deploying emailSendDefinition


lib/metadataTypes/EmailSendDefinition.js
```diff
@@ -110,7 +114,7 @@ class EmailSendDefinition extends MetadataType {
             // * this ignores r__assetMessage_Name on purpose as that is only unique per parent folder but useful during PR reviews
             metadata.Email.ID = cache.searchForField(
                 'asset',
-                metadata.r__assetMessage_Key,
+                metadata.r__assetMessage_Key.slice(0, Math.max(0, 36 - suffix.length)) + suffix,
```
JoernBerkefeld, line 117: try catch  around this and in the catch part you try cache.searchForField() again but without your new logic. if that also fails, so be it and we show the error that we already defined.

JoernBerkefeld: make sure to include the 2 delete statements in the try part, and in the catch part, after the cache.searchForField() part

JoernBerkefeld:
this PR should be revived now. the necessary fix for caching the global all subs list is now in develop branch & merged into this one @DannyCastroVentura

JoernBerkefeld:
during deploy tests, after fixing the obvious issues with SenderProfile, DeliveryProfile and SendClassification, I've been getting error 42117. Looking at https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/42000_42999_emailsenddefinition_object.html that error stands for **ESD_Invalid ObjectID** = **Invalid ObjectID**


could make sense during an update. but why during create?!

JoernBerkefeld:
now we get a 42116.... null value.

tested a lot but so far no idea whats wrong now.

PR #595, opened 2022-12-08
#594: handle SOAP errors happening during create and retrieve

### LINKED ISSUE #594: [BUG] output actual error message during CREATE and RETRIEVE via SOAP

### Current Behavior

we've updated this for UPDATE but forgot to do the same for CREATE. As a result we only show a generic error message and not the cause of the failure.

### Expected Behavior

_No response_

### Steps To Reproduce

3. Run '...'
4. See error...

### Version

4.1.12

### Environment

- OS:
- Node:
- npm:

### Participation

### Additional comments

_No response_

JoernBerkefeld:
# PR details

## What is the purpose of this pull request? 

- fixes #594

### DIFF
--- a/lib/metadataTypes/MetadataType.js
+++ b/lib/metadataTypes/MetadataType.js
@@ -614,7 +614,7 @@ class MetadataType {
             }
             return response;
         } catch (ex) {
-            this._handleSOAPErrors(ex, metadataEntry, 'creating', handleOutside);
+            this._handleSOAPErrors(ex, 'creating', metadataEntry, handleOutside);
             return {};
         }
     }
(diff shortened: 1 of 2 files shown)


lib/metadataTypes/MetadataType.js
```diff
@@ -721,11 +714,17 @@ class MetadataType {
...
-        );
+        let response;
+        try {
+            response = await this.client.soap.retrieveBulk(
+                this.definition.type,
+                fields,
+                requestParams
+            );
+        } catch (ex) {
+            this._handleSOAPErrors(ex, 'retrieving');
```
DougMidgley, line 725: is this really needed here? I would rather have all these errors bubble up where possible., rather than return empty objects, which has caused errors in the past.

JoernBerkefeld: happy to capture that rewrite in a dedicated ticket. for now, this is just adding a missing error handler. also, if we do want to equalize it, this method could later throw an error if we had the time to rewrite and test the surrounding logic - which I do not have right now

JoernBerkefeld: - added ticket #603

DougMidgley (approved):

PR #600, opened 2022-12-09
#599: in mcdev init use local user name/email if found, otherwise glo…

### LINKED ISSUE #599: [BUG] git user.name / user.email does not propose current local version

### Current Behavior

we seem to be only checking global settings

### Expected Behavior

we should let git try to read the local config as well. Otherwise, any local git config values will always be overwritten by `mcdev init` / `mcdev upgrade`

### Steps To Reproduce

1. run git config --local user.name "test"
2. run git config --global user.name "Your Name"
3. Run `mcdev init`
4. It will always propose "Your Name" to you

### Version

4.1.12

### Environment

- OS:
- Node:
- npm:

### Participation

### Additional comments

_No response_

JoernBerkefeld:
# PR details

- [x] Bugfix
- fixes #599 

replaced simple-git solution with git-cli solution

### DIFF
--- a/lib/util/init.git.js
+++ b/lib/util/init.git.js
@@ -252,30 +252,10 @@ const Init = {
      * @returns {Promise.<{'user.name': string, 'user.email': string}>} user.name and user.email
      */
     async _getGitConfigUser() {
-        const gitConfigs = await git.listConfig();
-        // remove local config
-        delete gitConfigs.values['.git/config'];
-        const result = {};
+        const name = Util.execSync('git', ['config', 'user.name'], true);
+        const email = Util.execSync('git', ['config', 'user.email'], true);
 
-        for (const file of Object.keys(gitConfigs.values)) {
-            if (gitConfigs.values[file]['user.name']) {
-                // sometimes the config has more than one entry per key - take the last one
-                result['user.name'] = Array.isArray(gitConfigs.values[file]['user.name'])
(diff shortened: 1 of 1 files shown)


lib/util/init.git.js
```diff
@@ -252,30 +252,10 @@ const Init = {
      * @returns {Promise.<{'user.name': string, 'user.email': string}>} user.name and user.email
      */
     async _getGitConfigUser() {
-        const gitConfigs = await git.listConfig();
-        // remove local config
-        delete gitConfigs.values['.git/config'];
-        const result = {};
+        const name = Util.execSync('git', ['config', 'user.name'], true);
```
DougMidgley, line 255: kinda confused what the objective is here? you are moving away from the helper lib to using native commands, but not sure why?

JoernBerkefeld: if we run the direct commands it automatically takes the local config first, then reads the local one. and it returns whatever value is in that config last. 
simple-git tried to be better but did it worse as a result. if there were multiple entries, it would return an array, even though that didn't make sense and git wouldn't care about that list but only use the last entry

JoernBerkefeld: alright. there was actually another command that resolved it like calling git via CLI does. i pushed the change and now we are back to using simple-git - but in a smarter way than before at least.

PR #629, opened 2023-01-13
Bugfix/628 journey builder triggeredsend not retrieved

### LINKED ISSUE #628: [BUG] journey builder triggeredSend not retrieved

### Current Behavior

we currently filter these entries with custom code.

### Expected Behavior

filtering should be done via the central filter option if needed. custom filters need to be removed

### Steps To Reproduce

3. Run '...'
4. See error...

### Version

4.2.0

### Environment

- OS:
- Node:
- npm:

### Participation

- [ ] I am willing to submit a pull request for this issue.

### Additional comments

_No response_

JoernBerkefeld:
# PR details

- fixes #628 

## Checklist

- [ ] test scripts updated


lib/metadataTypes/TriggeredSendDefinition.js
```diff
@@ -137,7 +123,11 @@ class TriggeredSendDefinition extends MetadataType {
         // remove IsPlatformObject, always has to be 'false'
         delete metadata.IsPlatformObject;
         // folder
-        super.setFolderPath(metadata);
+        try {
+            this.setFolderPath(metadata);
+        } catch {
+            return;
```
DougMidgley, line 129: any reason to change this logic? are we just ignoring errors here?

JoernBerkefeld: yes. reason is rather simple. by removing the filter on IsPlatformObject we get journey builder TSDs, but we also get shit ton of deleted & hidden stuff that is no longer visible in the GUI.
the CLI was flooded by warnings before I changed that to simply skip these entries.

ideal? no. it's really just a workaround for yet another quirk in SFMC :(

JoernBerkefeld: i should add that thats also the reason why I changed the other Util.loggers in this method from WARN to VERBOSE. These other 2 cases occur less often and one should be able to see it at least in the logs because those other entries do show up in the GUI. without a folder though, one couldn't see them anywhere anyways.

PR #631, opened 2023-01-13
Feature/627 add BASIC create+add new version+update+delete support for journeys interaction + #611 `delete` eventDefinition

### LINKED ISSUE #611: [FEATURE] support `delete` for eventDefinition
https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/deleteEventDefinition.html

/interaction/v1/eventDefinitions/key:{key}

### LINKED ISSUE #627: [FEATURE] add update/insert/delete support for journeys (interaction)
while supporting id rewrites is currently out of the question, basic deploy of already manually prepared jsons would already help.

- delete: https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/deleteInteractionById.html
- update: https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/putUpdateInteraction.html
- insert: https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/postCreateInteraction.html

general API overview: https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/jb-api-specification.html

JoernBerkefeld:
# PR details

- closes #627 
- closes #611 

- there is no parsing happening for INSERT / UPDATE. right now this is basically supposed to replace going via postman.

## Checklist

- [x] test scripts updated


lib/metadataTypes/definitions/Interaction.definition.js
```diff
@@ -128,9 +128,9 @@ module.exports = {
         },
         modifiedDate: {
             isCreateable: false,
-            isUpdateable: false,
+            isUpdateable: true,
```
DougMidgley, line 131: explain?

lib/metadataTypes/definitions/Interaction.definition.js
```diff
@@ -128,9 +128,9 @@ module.exports = {
...
             retrieving: true,
-            template: false,
+            template: true,
```
DougMidgley, line 133: explain?

lib/metadataTypes/definitions/Interaction.definition.js
```diff
@@ -388,15 +388,15 @@ module.exports = {
         },
         version: {
             isCreateable: false,
-            isUpdateable: false,
+            isUpdateable: true,
```
DougMidgley, line 391: explain?

lib/metadataTypes/definitions/Interaction.definition.js
```diff
@@ -388,15 +388,15 @@ module.exports = {
...
             retrieving: true,
-            template: false,
+            template: true,
```
DougMidgley, line 393: explain

lib/metadataTypes/MetadataType.js
```diff
@@ -624,12 +624,15 @@ class MetadataType {
      *
      * @param {TYPE.MetadataTypeItem} metadataEntry a single metadata Entry
      * @param {string} uri rest endpoint for PATCH
+     * @param {boolean} [usePut] some update requests require PUT instead of PATCH
      * @returns {Promise} Promise
      */
-    static async updateREST(metadataEntry, uri) {
+    static async updateREST(metadataEntry, uri, usePut) {
```
DougMidgley, line 630: from memory put is needed for a new version, whereas patch is used to update an existing draft -> please confirm behaviour on this first :)

lib/metadataTypes/MetadataType.js
```diff
@@ -822,6 +825,9 @@ class MetadataType {
                 }
             } else if (singleRetrieve) {
                 // some types will return a single item intead of an array if the key is supported by their api
+                if (singleRetrieve.startsWith('id:')) {
```
DougMidgley, line 828: this naming convention of adding id: will not always apply. some use ObjectId, some use other id (querydefinitionId). Just wondering if another command would not be more appropriate

lib/metadataTypes/Interaction.js
```diff
@@ -21,16 +21,59 @@ class Interaction extends MetadataType {
      * @returns {Promise.<TYPE.MetadataTypeMapObj>} Promise
      */
     static retrieve(retrieveDir, _, __, ___, key) {
+        let singleKey = '';
+        if (key) {
+            /* eslint-disable unicorn/prefer-ternary */
+            if (key.startsWith('id:')) {
+                // ! allow selecting journeys by ID because that's what users see in the URL
+                singleKey = '/' + encodeURIComponent(key.slice(3));
```
DougMidgley, line 29: id should not need encoding, what they see on the screen is what is needed by api.

DougMidgley (changes requested):

lib/metadataTypes/definitions/Interaction.definition.js
JoernBerkefeld: because the docs say you need to pass it in for updates:
"Although the modifiedDate may not be updated, you are required to pass current value that matches the database record."
>> https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/putUpdateInteraction.html

lib/metadataTypes/definitions/Interaction.definition.js
JoernBerkefeld: if it's needed for update, we need to include it for templates

lib/metadataTypes/Interaction.js
JoernBerkefeld: good point

lib/metadataTypes/definitions/Interaction.definition.js
JoernBerkefeld: thats the journey version you want to update

JoernBerkefeld: if we do not send it during an update, the API returns the error "Interaction not found" (did test this one in particular)

lib/metadataTypes/MetadataType.js
JoernBerkefeld: for journey updates, the docs state PUT is what you need. In my tests, patch didn't work at all while put did what the docs promised.
if you want to create a draft for an already running journey (and there can be only one draft), then you need to run the create (POST) call. if, however, there is a draft and you run PATCH, you get errors, if you run PUT for that draft, it updates it. rather strange but working

https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/putUpdateInteraction.html

lib/metadataTypes/MetadataType.js
JoernBerkefeld: the idea here is that you literally write `mcdev r cred/bu type id:asdadasdasd` to retrieve by id **asdadasdasd**. Meanwhile, the normal  `mcdev r cred/bu type dgfagfdsgdfgdf` would retrieve by key **dgfagfdsgdfgdf**.
Therefore, it's up to the user to add "id:" in front of an id to get this to work - if the type supports it. Not planning on adding that to any other type but journeys at this point. Therefore, this workaround seems better than adding a new command just for one type.

lib/index.js
```diff
@@ -354,10 +354,11 @@ class Mcdev {
      *
      * @param {string} businessUnit references credentials from properties.json
      * @param {string} type supported metadata type
-     * @param {string} customerKey Identifier of data extension
+     * @param {string} customerKey Identifier of metadata
+     * @param {number} [version] optional version of metadata
      * @returns {Promise.<void>} -
      */
-    static async deleteByKey(businessUnit, type, customerKey) {
+    static async deleteByKey(businessUnit, type, customerKey, version) {
```
DougMidgley, line 361: dont love that suddenly a journey concept becomes the default. would it not be better to use something like customerkey/version?

lib/metadataTypes/MetadataType.js
DougMidgley: i would rather you put flags if thats the case. using this makes it a naming convention not flag topic

lib/metadataTypes/Interaction.js
```diff
@@ -21,16 +24,265 @@ class Interaction extends MetadataType {
...
+        return super.createREST(metadata, '/interaction/v1/interactions/');
+    }
+
+    /**
+     * manages post retrieve steps
+     *
+     * @param {TYPE.MetadataTypeItem} metadata a single query
+     * @returns {TYPE.MetadataTypeItem} Array with one metadata object and one query string
+     */
+    static postRetrieveTasks(metadata) {
```
DougMidgley, line 114: so basically it is only partial given it will only basically work with email?

lib/metadataTypes/MetadataType.js
```diff
@@ -1531,9 +1570,10 @@ class MetadataType {
      *
      * @param {TYPE.BuObject} buObject references credentials
      * @param {string} customerKey Identifier of data extension
+     * @param {number} [version] optional version of metadata
      * @returns {boolean} deletion success status
      */
-    static deleteByKey(buObject, customerKey) {
+    static deleteByKey(buObject, customerKey, version) {
```
DougMidgley, line 1576: refer other comment on version

lib/metadataTypes/Interaction.js
JoernBerkefeld: it _works_ with everything but there are lots of IDs that might need to be resolved manually for cross-BU deployments.
As of now, it will allow updating & backing up journeys. I tried adding as many tests as possible for now, but it's certainly still lacking additional postRetrive & preDeploy processing to work as seamlessly as other types do.

I would call this a beta release officially which will be improved over time. Luckily, journeys often use keys already, making custom r__ or c__ fields irrelevant for the most part

lib/index.js
JoernBerkefeld: it's an optional field after all - the type-classes have to implement using the version (or not)
I see your point, arent there any other types that might be able to use versions? :D

lib/metadataTypes/MetadataType.js
JoernBerkefeld: can you give an example? I don't understand

JoernBerkefeld:
wiki updated:
- journeys section: https://github.com/Accenture/sfmc-devtools/wiki/8.-Metadata-specific-settings-&-options#journeys-interaction
- supported metadata: https://github.com/Accenture/sfmc-devtools/wiki/5.-Metadata-Type-Support
- delete-support: https://github.com/Accenture/sfmc-devtools/wiki/6.b-~-Standard-Commands#delete

lib/index.js
DougMidgley: as discussed - key/version construct

lib/metadataTypes/Interaction.js
DougMidgley: ensure that there is a real comment to explain limited support like "IMPORTANT: this is for beta release supporting only native external keys OR triggered sends, not all activities are supported"
AND ADD A LOG/WARN to user

lib/metadataTypes/MetadataType.js
DougMidgley: ensure at least in docs that this is clear what it will actually work with.

DougMidgley (changes requested):

lib/metadataTypes/Interaction.js
JoernBerkefeld: done :)

lib/index.js
JoernBerkefeld: done

lib/metadataTypes/MetadataType.js
JoernBerkefeld: fixed / reverted

JoernBerkefeld:
will go ahead with the merge after applying the agreed-upon fixes, @DougMidgley

JoernBerkefeld:
wiki updated again

PR #639, opened 2023-01-18
#638: ensure token requests are not counted

### LINKED ISSUE #638: [TASK] ensure test cases dont fail on api-request-count due to tokens
for whatever reason resetting tokens does not seem to work in between test cases. 
as a result, only the first test class that gets executed (currently "dataExtension.test") needs to get a token while the others don't. This means that the first test in a class always fails when trying to execute it without first running dataExtension tests.

JoernBerkefeld:
# PR details

- [x] Technical debt removed

- fixes #638 

## Checklist

- [x] test scripts updated

### DIFF
--- a/test/dataExtension.test.js
+++ b/test/dataExtension.test.js
@@ -29,9 +29,9 @@ describe('dataExtension', () => {
                 'returned metadata was not equal expected'
             );
             assert.equal(
-                Object.values(testUtils.getAPIHistory()).flat().length,
-                6,
-                'Unexpected number of requests made'
+                testUtils.getAPIHistoryLength(),
+                5,
+                'Unexpected number of requests made. Run testUtils.getAPIHistoryDebug() to see the requests'
             );
             return;
         });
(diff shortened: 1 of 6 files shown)


test/dataExtension.test.js
```diff
@@ -29,9 +29,9 @@ describe('dataExtension', () => {
                 'returned metadata was not equal expected'
             );
             assert.equal(
-                Object.values(testUtils.getAPIHistory()).flat().length,
-                6,
-                'Unexpected number of requests made'
+                testUtils.getAPIHistoryLength(),
+                5,
+                'Unexpected number of requests made. Run testUtils.getAPIHistoryDebug() to see the requests'
```
DougMidgley, line 34: how does one "run" this?

DougMidgley (changes requested):
If you are confident that this is the right thing to do then lets go ahead, if its a workaround, id rather see a real solution to it (ie. why its not resetting)

test/dataExtension.test.js
JoernBerkefeld: by adding testUtils.getAPIHistoryDebug() to your testing code?
I didn't want that to be in the code in some static fashion because in most situations it's clear what went wrong. But if it's not, then we now have that extra method for more debugging.

JoernBerkefeld: on your workaround question: well, we are calling the method provided by the external library but it doesn't seem to work. It caches the token despite that. From my POV the issue needs to be resolved by the lib owner eventually. Until then, I think this is a perfectly fine solution for the issue

PR #721, opened 2023-02-07
#706: refresh triggeredSends on email update (deploy)

### LINKED ISSUE #706: [FEATURE] refresh triggeredSends on email update
add option --refresh to deploy command which calls a helper function that finds relevant triggeredSends for the assets that are getting deployed - and then runs the refresh command for them.

image

JoernBerkefeld:
# PR details

- closes #706 

this adds a new option to `deploy` which calls `refresh` under the hood to update the right triggeredSends after updating assets. not implemented for other types!

_Example:_
`mcdev deploy --refresh`

## Checklist

- n/a test scripts updated
- [x] Wiki updated (if applicable) https://github.com/Accenture/sfmc-devtools/issues/722


lib/metadataTypes/Asset.js
```diff
@@ -495,6 +495,78 @@ class Asset extends MetadataType {
...
+    /**
+     * Gets executed after deployment of metadata type
+     *
+     * @param {TYPE.MetadataTypeMap} metadata metadata mapped by their keyField
+     * @param {TYPE.MetadataTypeMap} _ originalMetadata to be updated (contains additioanl fields)
+     * @param {{created: number, updated: number}} createdUpdated counter representing successful creates/updates
+     * @param {boolean} [isRefresh] optional flag to indicate that triggeredSend should be refreshed after deployment of assets
+     * @returns {Promise.<void>} -
+     */
+    static async postDeployTasks(metadata, _, createdUpdated, isRefresh) {
```
DougMidgley, line 507: things like this flags, ideally, shouldnt be passed, but be available via a context object to avoid that we pass 1000 things via method

lib/metadataTypes/Automation.js
```diff
@@ -230,12 +230,11 @@ class Automation extends MetadataType {
      * @param {TYPE.AutomationMap} metadata metadata mapped by their keyField
      * @param {string} targetBU name/shorthand of target businessUnit for mapping
      * @param {string} retrieveDir directory where metadata after deploy should be saved
+     * @param {boolean} [isRefresh] optional flag - so far not used by automation
      * @returns {Promise.<TYPE.AutomationMap>} Promise
      */
-    static async deploy(metadata, targetBU, retrieveDir) {
```
DougMidgley, line 235: sure this doesnt break something?

JoernBerkefeld: at least in my tests things kept working afterwards. 
i did a find-in-folder for postDeployTasks, upsert and deploy to find all occurrences across classes

lib/metadataTypes/Asset.js
JoernBerkefeld: I do agree when it comes to global flags. But this one is limited to this command and hence I don't see the point to clutter a central class, e.g. util, with this variable.
This is a parameter that is used specifically for this case and hence it should continue to be a function parameter of this function

DougMidgley: your comment is a little contradictory. You want to add this to all common classes because it is needed here? having a context class is exactly the kind of thing that should go via a context class to ensure you DONT have to pollute all the other methods with additional parameters. generally, beyond 3 parameters suggests something else is not well designed

JoernBerkefeld: We will actually have a second use case: attribute groups on child bus when deploying shared data extensions.

Currently working on getting these fully resolved on another branch

Issue #722, opened 2023-02-08
[TASK] update documentation for 5.0.0 release


JoernBerkefeld:
- #829 / PR #858

https://github.com/Accenture/sfmc-devtools/wiki/02.-Getting-Started/_edit

### Install Node.js with npm

1. To check if it is already installed, at the OS command prompt, type: `node --version`
    - If this command reports Node version 15.0.0 or later, you’re done—proceed to the next installation. If the reported version is earlier than 15.0.0, continue to step 2.
    - If you get a “command not found” error message, continue to step 2.
2. In a web browser, go to [nodejs.org](https://nodejs.org)
3. Download and run the latest **LTS** installer for your operating system.
4. When the installer finishes, try step 1 again. If it fails, please restart your terminal. If it still does not work, reboot your computer and try the version check then.

- #139 / PR #749 + PR #814
- #817 / PR #857

https://github.com/Accenture/sfmc-devtools/wiki/8.-Metadata-specific-settings-&-options

### Retrieve & delete by ID

Journey is the only type that supports ids for `retrieve` and `delete`. The reason is that the key is not exposed in the interface. Meanwhile, the ID is directly visible in the URL (together with the version). To use this, prefix your ID with `id:` like in the following example, in which 'afjdskfsafkldsafklf' is the ID you found in the URL:

```bash
mcdev retrieve cred/bu journey id:afjdskfsafkldsafklf
```

## Users (`user`)

You can create/update most fields in the `user` JSONs, **except for User Permissions**. Roles and Business Units can be assigned and removed at will simply by adding or removing the role name or the MID, respectively.

### Roles associated with Business Units

While the user interface of SFMC allows setting up roles that will be automatically assigned in the background to any user with access to that BU, we cannot access that information via API at this point (or at least we do not know how).

### Roles associated with users only for a specific Business Unit

SFMC DevTools cannot show which roles were assigned to users just for given Business Units, nor can it update such an assignment.

### Standard "Marketing Cloud *" roles

Due to a bug in SFMC's APIs, whenever we try to update a user that has one of these standard Marketing Cloud roles assigned, these roles will be unassigned. This means if you had such a role assigned to a user, and then you try to update that user, even if it does not concern the roles, these default roles will be removed afterwards. If, on the other hand, you try to add such a role, this attempt will be ignored.

[...]

PR #825, opened 2023-03-27
Feature/782 support UPDATE and DELETE for `mobileKeyword` and extract responseMessage to AMP-file

### LINKED ISSUE #782: [FEATURE] support UPDATE and DELETE for `mobileKeyword` and extract responseMessage to AMP-file
this can be a 1-liner but it can also contain code and line breaks --> should be extracted into separate file, ideally and ampscript file.
image

JoernBerkefeld:
# PR details

- closes #782 

## Checklist

- [x] test scripts updated
- [x] Wiki updated (for DELETE) via #722


lib/metadataTypes/definitions/MobileKeyword.definition.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
     bodyIteratorField: 'entry',
     dependencies: ['mobileCode'],
     hasExtended: false,
-    idField: 'decodedId',
+    idField: 'id',
```
DougMidgley, line 5: i put it with decoded because it is used in other references - not sure if you triple checked this?

lib/metadataTypes/definitions/MobileKeyword.definition.js
```diff
@@ -16,14 +16,15 @@ module.exports = {
...
-            isUpdateable: false,
-            retrieving: false,
+            isUpdateable: true,
+            retrieving: true,
             template: false,
         },
         keyword: {
             isCreateable: true,
-            isUpdateable: false,
+            isUpdateable: true,
```
DougMidgley, line 27: is it updateable?

lib/metadataTypes/definitions/MobileKeyword.definition.js
JoernBerkefeld: if you can tell me WHERE then sure thing. in my tests, `id` was used in journey

lib/metadataTypes/definitions/MobileKeyword.definition.js
JoernBerkefeld: it needed to be included during updates or else the API returned an error. no, you cannot change it.

lib/metadataTypes/definitions/MobileKeyword.definition.js
DougMidgley: just somewhere in memory is all. If you are ok, then lets move it forward iwth ID

DougMidgley (approved):

PR #919, opened 2023-05-21
Feature/730 feature execute query command

### LINKED ISSUE #730: [FEATURE] `execute` query command
The Return is very basic with just “OK” or “Error”.

POST /automation/v1/queries/{{queryID}}/actions/start/

should support 
- `*` for BUs
- marketlist for BUs
- multiple query keys

later, the same command can be expanded for starting journeys, triggeredSends, ...

phjulia:
# PR details

- closes #730 
Added a function to execute a query/queries. Execute all queries on all BUs if * is provided instead of a BU and no keys

## Checklist

- [x] test scripts updated


lib/metadataTypes/Query.js
```diff
@@ -50,6 +50,32 @@ class Query extends MetadataType {
...
+    static async execute(keyArr) {
+        if (keyArr[0] == null) {
+            keyArr = await this.findItemsToExecute();
+        }
+        for (const key of keyArr) {
+            const uri = `/automation/v1/queries/${key}/actions/start/`;
+            Util.logger.info(`Starting item ${key}`);
+            super.executeREST(uri);
+        }
+        return 'OK';
```
JoernBerkefeld, line 66: - [ ] replace static return with `Promise.all()` logic

example from `TriggeredSend.js > refresh()`:

```js
        const refreshList = [];
        Util.logger.info(`Refreshing ${keyArr.length} ${this.definition.typeName}...`);
        Util.logger.debug(`Refreshing keys: ${keyArr.join(', ')}`);
        for (const key of keyArr) {
            refreshList.push(this._refreshItem(key, checkKey));
        }
        const successCounter = (await Promise.all(refreshList)).filter(Boolean).length;
        Util.logger.info(`Refreshed ${successCounter} of ${keyArr.length}`);
        return successCounter === keyArr.length;

```

I appended the return-line which does not exist in the refresh function. That was overlooked when designing the refresh function and should also be added there in a future release. Alternatively, one could imagine to return for example an object with success & error counters or even a more detailed object that clearly lists which ones worked and which ones didn't... given that we don't have such a requirement, a simple "all worked" or "at least one failed" solution like the suggested one above will likely do.

lib/metadataTypes/MetadataType.js
```diff
@@ -1070,6 +1070,24 @@ class MetadataType {
...
+     */
+    static async executeREST(uri) {
+        try {
+            const response = await this.client.rest.post(uri, {}); //payload is empty for this request
+            if (response == 'OK') {
+                Util.logger.info(`\n               :: Executed ${this.definition.type}`);
+            } else {
+                Util.logger.info(`${this.definition.type} was not executed`);
+            }
+            return response;
```
JoernBerkefeld, line 1086: good

lib/metadataTypes/MetadataType.js
```diff
@@ -1070,6 +1070,24 @@ class MetadataType {
             type: this.definition.type,
         };
     }
+    /**
+     * Used to execute a query/automation etc.
+     *
+     * @param {string} uri REST endpoint where the POST request should be sent
```
JoernBerkefeld, line 1076: - [ ] define return value with `@returns {Promise.<string>} your defintion here` line

lib/metadataTypes/Query.js
```diff
@@ -50,6 +50,32 @@ class Query extends MetadataType {
             key
         );
     }
+    /**
+     * a function to start query execution via API
+     * @param {string[]} keyArr customerkey of the metadata
```
JoernBerkefeld, line 55: - [ ] define return value with `@returns {Promise.<string>} your defintion here` line

lib/metadataTypes/MetadataType.js
```diff
@@ -1070,6 +1070,24 @@ class MetadataType {
...
+     */
+    static async executeREST(uri) {
+        try {
+            const response = await this.client.rest.post(uri, {}); //payload is empty for this request
+            if (response == 'OK') {
```
JoernBerkefeld, line 1081:
```suggestion
            if (response === 'OK') {
```

test/type.query.test.js
```diff
@@ -277,4 +277,22 @@ describe('type: query', () => {
             return;
         });
     });
+    describe('Execute ================', () => {
+        it('Should start executing queries', async () => {
+            const execute = await handler.execute('testInstance/testBU', ['query']);
+            assert.equal(process.exitCode, false, 'execute should not have thrown an error');
+            assert.equal(execute, true, 'queries were supposed to be executed');
+            return;
+        });
```
JoernBerkefeld, line 286: this test will need to be removed or changed to execute 2+ specific queries - and test for a different return value

test/type.query.test.js
```diff
@@ -277,4 +277,22 @@ describe('type: query', () => {
             return;
         });
     });
+    describe('Execute ================', () => {
+        it('Should start executing queries', async () => {
+            const execute = await handler.execute('testInstance/testBU', ['query']);
```
JoernBerkefeld, line 282: you seem to have an async issue because both tests failed due to missing api-response files:

- [ ] add response files: `test\resources\9999999\automation\v1\queries\<KEY>\actions\start\post-response.json`

image


please make sure you always run the test cases yourself locally and actually check the output of the cases that you add / update or that might be affected by your change in any way to ensure you don't miss things like this async issue. Here, the test case finishes before the mock-API call is executed and hence mocha does not pick up on the issue.

you can do so by running `npm run test`, or by hitting CTRL+SHIFT+T (might ask you to pick default test task once) or, if you only want to run one test class, by directly naming it like this (with the example of query test class):

`npx mocha test/type.query.test.js`

JoernBerkefeld: another way of running tests, though the test log is then in the OUTPUT tab, not in the TERMINAL tab - which doesn't have color coding - is to use the VSCode extension [Mocha Test Explorer](https://marketplace.visualstudio.com/items?itemName=hbenl.vscode-mocha-test-adapter). I use that regularly as well

image

lib/index.js
```diff
@@ -714,14 +717,20 @@ class Mcdev {
                 Util.logger.info(`\n               :: Executing the entity on all BUs for ${cred}`);
 
                 for (const bu in properties.credentials[cred].businessUnits) {
-                    await this._executeBU(cred, bu, selectedTypesArr, keys);
-                    counter_credBu++;
+                    result = await this._executeBU(cred, bu, selectedTypesArr, keys);
```
JoernBerkefeld, line 720: this overwrites previous failures. please collect failures AND success in separate variables, then check if you found any errors

JoernBerkefeld (changes requested):

JoernBerkefeld:
good to merge now!
let me sort if we should go for 5.1.0 already and then merge it eventually

JoernBerkefeld (approved):

PR #992, opened 2023-06-16
bugfix/988 read refresh from options

phjulia:
# PR details

Read refresh from the options

...

## Checklist

- N/A test scripts updated
- N/A Wiki updated (if applicable)


lib/metadataTypes/TriggeredSend.js
```diff
@@ -127,17 +127,16 @@ class TriggeredSend extends MetadataType {
...
-        // email
+        // try {
+        //     this.setFolderPath(metadata);
+        // } catch {
+        //     Util.logger.verbose(
+        //         ` ☇ skipping ${this.definition.typeName} '${metadata.Name}'/'${metadata.CustomerKey}': Could not find folder.`
+        //     );
+        //     // do not save this TSD because it would not be visible in the user interface
+        //     return;
+        // }
```
phjulia, line 138: @JoernBerkefeld returned metadata is undefined and refresh does not work if we have this piece of code executed. Do we need this for refresh? I don't understand why do we need to set folder if this is not a retrieve?

JoernBerkefeld: Hm, I'm having a hard time checking this on my phone but: this method is only executed during a retrieve. Not during caching.

Therefore, either our refresh logic retrieves (instead of caches) for no reason or we needed the download including saving to disk (aka retrieve) for the refresh somehow.

JoernBerkefeld: We execute it here to validate the TS:

https://github.com/Accenture/sfmc-devtools/blob/4093f49bfe64f026eaee372dce83ac5f02768789/lib/metadataTypes/TriggeredSend.js#L277

JoernBerkefeld: And here we should check if type==="folder" or cache...

Otherwise the necessary folder subtypes for TS are not cached

phjulia: @JoernBerkefeld somehow an error was thrown in searchForField() function that was not caught by try catch on lines 103-119. I've added a try catch and that helped to avoid the problem, refresh completes successfully. 
I don't entirely understand what kind of folders are we talking about when it comes to TriggeredSends, maybe these ones (the corresponding triggered send is located in the Journey sends -> journey)? Is there a problem somewhere with caching that those folders aren't cached?
image
This is what it looks like when an outer try catch has been added:
image

lib/metadataTypes/TriggeredSend.js
```diff
@@ -136,8 +135,7 @@ class TriggeredSend extends MetadataType {
             // do not save this TSD because it would not be visible in the user interface
             return;
         }
-
-        // email
+        // // email
```
JoernBerkefeld, line 138: revert please
```suggestion
        
        // email
```

lib/metadataTypes/TriggeredSend.js
JoernBerkefeld: pushed it forward run without errors
image

JoernBerkefeld (changes requested):
could you add a test class for asset to ensure this keeps working?

JoernBerkefeld:
asset test class shall be done as part of:
- #1002

that ensures we can push out this bugfix faster

JoernBerkefeld (approved):

PR #1006, opened 2023-06-27
Feature/987 instead of using key allow using filter similar to sqls like logic

### LINKED ISSUE #987: [FEATURE] allow using `--like` option similar to SQL's `LIKE` logic
We often have to update a large subset of metadata, e.g. disable certain users but have issues finding the right ones.

**Should be supported by:**
- [x] retrieve
- [ ] execute
- [ ] refresh

**out of scope:**
- [ ] deploy
- [ ] buildTemplate
- [ ] buildDefinition
- [ ] buildDefinitionBulk

why? because having support in retrieve alone allows us to run deploy and buildX based on the result already.

**logic:**
- Filter should support finding strings or check json fields with LIKE % to get partial matches

**solution:**
- `--like.field "%abc"` --> Match a string that ends with abc, similar to LIKE '%abc'
- `--like.field "abc%"` --> Match a string that starts with abc, similar to LIKE 'abc%'

https://stackoverflow.com/a/18418386/818732:

```js

function stringLike(testString, search) {
    if (typeof search !== 'string' || this === null) {return false; }
    // Remove special chars
    search = search.replace(new RegExp("([\\.\\\\\\+\\*\\?\\[\\^\\]\\$\\(\\)\\{\\}\\=\\!\\<\\>\\|\\:\\-])", "g"), "\\$1");
    // Replace % and _ with equivalent regex
    search = search.replace(/%/g, '.*').replace(/_/g, '.');
    // Check matches
    return RegExp('^' + search + '$', 'gi').test(testString);
}

const url = 'http://www.mydomain.com/page1.aspx';
console.log(stringLike(url, '%mydomain.com/page_.asp%')); // true
```

https://www.w3schools.com/sql/sql_like.asp:
The LIKE operator is used in a WHERE clause to search for a specified pattern in a column.

There are two wildcards often used in conjunction with the LIKE operator:

 The percent sign (%) represents zero, one, or multiple characters
 The underscore sign (_) represents one, single character

LIKE Operator | Description
-- | --
--like.Name "a%" | Finds any values that start with "a"
--like.key "%a" | Finds any values that end with "a"
--like.name "%or%" | Finds any values that have "or" in any position
--like.CustomerKey "_r%" | Finds any values that have "r" in the second position
--like.Description "a_%" | Finds any values that start with "a" and are at least 2 characters in length
--like.customerKey "a__%" | Finds any values that start with "a" and are at least 3 characters in length
--like.steps.activities.name "a%o" | Finds any values that start with "a" and ends with "o"

JoernBerkefeld:
# PR details

- closes #987 

## Further details (optional)

...

## Checklist

- [ ] test scripts updated

### DIFF
--- a/docs/dist/documentation.md
+++ b/docs/dist/documentation.md
@@ -5880,6 +5880,8 @@ CLI entry for SFMC DevTools
     * [.getKeysString(keyArr, [isId])](#Util.getKeysString) ⇒ <code>string</code>
     * [.sleep(ms)](#Util.sleep) ⇒ <code>Promise.&lt;void&gt;</code>
     * [.getSsjs(code)](#Util.getSsjs) ⇒ <code>string</code>
+    * [.stringLike(testString, search)](#Util.stringLike) ⇒ <code>boolean</code>
+    * [.fieldsLike(metadata)](#Util.fieldsLike) ⇒ <code>boolean</code>
 
 <a name="Util.skipInteraction"></a>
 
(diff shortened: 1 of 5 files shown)


JoernBerkefeld:
need to come up with a good test for this.

Issue #1007, opened 2023-06-29
[TASK] Documentation for 5.2


JoernBerkefeld:
- #982 / PR #986
- #870 / PR #1019 
https://github.com/Accenture/sfmc-devtools/wiki/06.b-~-Standard-Commands/_edit

### deploy

_Command_: `mcdev deploy [business unit] [metadata type] [metadata key] [--fromRetrieve] [--refresh] [--changeKeyValue=yourNewKey] [--changeKeyField=otherFieldInJson] [--execute]`

**[...]**

**deploy & execute:**

This special combination will execute/start items after creating or updating them on the server.

_Example:_

```
mcdev deploy MyProject/DEV query --execute
mcdev deploy MyProject/DEV query "key1,key2" --execute
mcdev deploy MyProject/DEV query "key1" --fromRetrieve --execute

mcdev deploy MyProject/DEV automation --execute
mcdev deploy MyProject/DEV automation "key1,key2" --execute
mcdev deploy MyProject/DEV automation "key1" --fromRetrieve --execute
```

**Currently supported types for deploy & execute:**

| Name | CLI Argument | Effect |
| :-- | :-- | :-- |
| Automation | automation | Schedules automation according to already existing (but paused) schedule |
| Query | query | Starts query execution |

- #987 / PR #1007

https://github.com/Accenture/sfmc-devtools/wiki/06.b-~-Standard-Commands/_edit

### retrieve

[...]

**retrieve specific type and key via --like option:**

_Example:_

```bash
mcdev retrieve MyProject/DEV dataExtension --like.key "key1%"
mcdev retrieve MyProject/DEV asset --like.steps.activities.name "a%o"
mcdev retrieve MyProject/DEV asset --like.key "key1%" --like.steps.activities.name "a%o"
```
How does it work? There are two wildcards often used in conjunction with the LIKE operator:

-  The percent sign (%) represents zero, one, or multiple characters
 - The underscore sign (_) represents one, single character

LIKE Operator | Description
-- | --
--like.Name "a%" | Finds any values that start with "a"
--like.key "%a" | Finds any values that end with "a"
--like.name "%or%" | Finds any values that have "or" in any position
--like.CustomerKey "_r%" | Finds any values that have "r" in the second position
--like.Description "a_%" | Finds any values that start with "a" and are at least 2 characters in length
--like.customerKey "a__%" | Finds any values that start with "a" and are at least 3 characters in length
[...]

PR #1022, opened 2023-07-10
Feature/38 add fixkeys command

### LINKED ISSUE #1234: [TASK] investigate minimum scope / permissions required to run mcdev

phjulia:
# PR details

- closes #1234

## Further details (optional)

...

## Checklist

- [+] I have performed a self-review of my own code
- [+] I have commented my code, particularly in hard-to-understand areas
- [+] test scripts updated
- [+] Wiki updated (if applicable)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -859,11 +859,11 @@ class Mcdev {
      * @param {string} bu name of BU
      * @param {TYPE.SupportedMetadataTypes[]} [selectedTypesArr] limit execution to given metadata type
      * @param {string[]} keyArr customerkey of the metadata
-     * @returns {Promise.<boolean>} true if all items were executed, false otherwise
+     * @returns {Promise.<Object.<string,TYPE.MultiMetadataTypeMap>>} deployed metadata per BU (first key: bu name, second key: metadata type)
      */
-    static async fixKeys(bu, selectedTypesArr, keyArr) {
+    static async fixKeys(businessUnit, type, keys) {
         Util.startLogger();
-        Util.logger.info('mcdev:: FixKeys');
+        Util.logger.info('mcdev:: fixKeys');
         const properties = await config.getProperties();
         let counter_credBu = 0;
         let counter_failed = 0;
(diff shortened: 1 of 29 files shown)


lib/index.js
```diff
@@ -977,6 +977,203 @@ class Mcdev {
...
+        //         bu = buObject.businessUnit;
+        //     }
+        // }
+        // update keys on a BU
+        // fixKeyResult[]
+        // if (fixKeyResult.push(await this._fixKeysBU(cred, bu, type, keys))) {
+        //     counter_credBu++;
+        // } else {
+        //     counter_failed++;
+        // }
```
JoernBerkefeld: please remove

lib/index.js
```diff
@@ -977,6 +977,203 @@ class Mcdev {
...
+
+        let numItemsToUpdate;
+        let deployed;
+        0;
```
JoernBerkefeld: remove please
```suggestion
```

lib/index.js
```diff
@@ -977,6 +977,203 @@ class Mcdev {
...
+            }
+            Util.OPTIONS.changeKeyField = MetadataTypeDefinitions[type].nameField;
+            properties.directories.deploy = properties.directories.retrieve;
+            deployed = await Deployer._deployBU(cred, bu, properties, [type], keysForDeploy, true);
+            // TODO find types that are dependent on this type by iterating over THEIR dependencies
+            const dependentTypes = await this.#getDependentMetadata(type);
+            await retriever.retrieve(dependentTypes, null, null, false);
+            Util.OPTIONS.changeKeyField = false;
+            await Deployer._deployBU(cred, bu, properties, dependentTypes, null, true);
+            // TODO update other types that are dependent on this type.
```
JoernBerkefeld: lets do that TODO in a new ticket

lib/index.js
```diff
@@ -977,6 +977,203 @@ class Mcdev {
...
+    //         const dependentTypes = await this.#getDependentMetadata(type);
+    //         await retriever.retrieve(dependentTypes, null, null, false);
+    //         Util.OPTIONS.changeKeyField = false;
+    //         await Deployer._deployBU(cred, bu, properties, dependentTypes, null, true);
+    //         // TODO update other types that are dependent on this type.
+    //     } catch (ex) {
+    //         Util.logger.errorStack(ex, 'mcdev.fixKeys failed');
+    //     }
+    //     return Object.keys(Object.values(deployed)[0]);
+    // }
```
JoernBerkefeld: remove please 

```suggestion
```

lib/index.js
```diff
@@ -977,6 +977,103 @@ class Mcdev {
...
+            // merge results if multiple keys were provided
+            if (keys?.length > 1) {
+                const base = typeArr[0];
+                for (let i = 1; i < typeArr.length; i++) {
+                    // merge all items into the first array
+                    Object.assign(base, typeArr[i]);
+                }
+                // kick out all but first entry
+                typeArr.length = 1;
+            }
```
JoernBerkefeld, line 1041: this will likely have to be removed (should be fixed on develop branch)

```suggestion
```

JoernBerkefeld (changes requested):
please check comments and fix failing tests

lib/index.js
```diff
@@ -977,6 +977,103 @@ class Mcdev {
...
+            numItemsToUpdate = keysForDeploy.length;
+            if (numItemsToUpdate < 1) {
+                Util.logger.info(`No items to update. Exiting`);
+                return true;
+            }
+            Util.OPTIONS.changeKeyField = MetadataTypeDefinitions[type].nameField;
+            properties.directories.deploy = properties.directories.retrieve;
+            deployed = await Deployer._deployBU(cred, bu, properties, [type], keysForDeploy, true);
+            const dependentTypes = await this.#getDependentMetadata(type);
+            await retriever.retrieve(dependentTypes, null, null, false);
```
JoernBerkefeld, line 1052: instead of retrieving all types here, lets instead print a warning to retrieve these these types. Otherwise we might be overwriting work that was prepared offline. It's might be unexpected to users that we would touch types that were not specified in the command.

lib/index.js
```diff
@@ -977,6 +977,103 @@ class Mcdev {
 
         return keyArr;
     }
+    /**
+     * Updates the key to match the name field
+     *
+     * @param {string} businessUnit name of BU
+     * @param {TYPE.SupportedMetadataTypes} type limit execution to given metadata type
+     * @param {string[]} [keys] customerkey of the metadata
+     * @returns {Array} array of fixedKeys
```
JoernBerkefeld, line 986:
```suggestion
     * @returns {string[]} list of fixedKeys
```

lib/index.js
```diff
@@ -977,6 +977,103 @@ class Mcdev {
...
+        Util.logger.info(`\n               :: Done\n`);
+        return Object.keys(Object.values(deployed)[0]);
+    }
+    /**
+     *helper for {@link Mcdev.fixKeys}. Retrieve dependent metadata
+     *
+     * @private
+     *
+     * @param {string} type type of the metadata passed as a parameter to fixKeys function
+     * @returns {Array} array of types that depend on the given type
```
JoernBerkefeld, line 1065:
```suggestion
     * @returns {string[]} array of types that depend on the given type
```

JoernBerkefeld (changes requested):

test/type.query.test.js
```diff
@@ -207,6 +223,69 @@ describe('type: query', () => {
...
+            expect(
+                file(testUtils.getActualFile('testExisting_query_fixedKeys', 'query', 'sql'))
+            ).to.equal(file(testUtils.getExpectedFile('9999999', 'query', 'patch_fixKeys', 'sql')));
+            // check number of API calls
+            assert.equal(
+                testUtils.getAPIHistoryLength(),
+                16,
+                'Unexpected number of requests made. Run testUtils.logAPIHistoryDebug() to see the requests'
+            );
+            properties.directories.deploy = deployDir;
```
phjulia, line 285: @JoernBerkefeld 
I set deploy directory for fixKeys to retrieve dir, it causes problems in deploy script test. Is there anyway to reset execution context for when fixKeys test is done? This line is a workaround to avoid tests failing (see line 1078 in index.js)

JoernBerkefeld: why dont you use the option fromRetrieve before running fixKeys()?


add this before your handler.fixKeys(...) call in your test class: 

```js
handler.setOptions({fromRetrieve:true});
```

phjulia: > ```js
> handler.setOptions({fromRetrieve:true});
> ```

The problem is in re-setting the deploy directory. I deploy from retrieve directory so I set deploy directory to retrieve. In tests before I attempt to deploy a script in test.script.test I need to set deploy directory back to deploy because tests fail. 
For example if I were to use Deployer.deploy() instead of deployBU in fixKeys() it will set deploy directory to retrieve, but in tests I still need to set it back to deploy/. 
image

JoernBerkefeld: hi, actually my advice was only half correct:

using fromRetrieve:true IS the way to go, but only if you use the main Developer.develop method.
I removed that parameter that you screenshotted to make it consistent with similar options. Now, the methods look for Util.OPTIONS.fromRetrieve instead.

with that said, using handler.... is obviously wrong because that would only work for the test. instead, you'd have to manually set the option in fixKeys in the same way: `this.setOptions({fromRetrieve:true});`


if you need to go with deployBU() instead, then you'll have to adapt the hack that I added to your method:


before:
```js
        const deployDirBak = properties.directories.deploy;
        if (Util.OPTIONS.fromRetrieve) {
            properties.directories.deploy = properties.directories.retrieve;
        }
```

and after:
```js
        if (Util.OPTIONS.fromRetrieve) {
            properties.directories.deploy = deployDirBak;
        }
```

JoernBerkefeld: PS: please merge develop again ;-)

JoernBerkefeld (changes requested):
code is now good! 
only waiting for documentation proposal

JoernBerkefeld:
wiki entry polished and copied to #1007 for publishing

JoernBerkefeld (approved):

PR #1417, opened 2024-07-05
Bug/1406 creating events fails if dekey is set alongside schema

### LINKED ISSUE #1406: [BUG] creating events fails if deKey is set alongside schema

### Current Behavior

in our scenario, schema does NOT have a de-key set but the generic de-key is set outside of the schema.
pre-deploy steps look for the dataExtension and because we are trying to create the API event, that DE also doesn't exist

### Expected Behavior

we want SFMC to auto-create it for us based on the schema.
for that, the schema needs to have the DE key set so that it knows how to name the DE.
also, the preDeploy step needs to accept that it cant find the DE for CREATE only

### Version

7.0.3

JoernBerkefeld:
# PR details

- closes #1406 

## Further details (optional)

implemented a fallback routine:
1. search for specified DE on server and pick that if found
2. else, if schema is provided in event json, let eventDefinition API create a DE for us
3. rename and rekey the auto-created DE 

## Checklist

- [x] test scripts updated
- [ ] Wiki updated (**TODO**)


JoernBerkefeld:
# Documentation 
> https://github.com/Accenture/sfmc-devtools/wiki/08.-Metadata-specific-settings-%26-options/_edit

## Event (`event`)

PR #2284, opened 2025-05-21
#2296: Don't try to load prettier config if formatOnSave is false

justinjxzhang:
# PR details
I noticed while trying to `retrieve` `query` I was getting an error around `prettier` configurations, even though I had `formatOnSave` set to `false`.

While the retrieval still continued, the program exits with code `1` (due to the `Util.logger.error` function calling `Util.signalFatalError` which sets `process.exitCode = 1`) so this caused needing a bit of a workaround to get it working programatically inside another application.

If the `formatOnSave` option is `false` then the `initPrettier` function will early exit without trying to find the `prettierrc` config.

## Further details (optional)

I tried to add tests around the `initPrettier` function but couldn't see a straightforward way. I wanted to mock the `Util` module in order to check that `Util.logger.error` wasn't called, but it appears that module mocking is only supported as an experimental feature in `node:test` at the moment. I didn't want to go too far overboard with re-structuring things (especially in a first contribution), and given the tests appear to primarily be around the types or an end-to-end process, I thought it might squeak by.

...

## Checklist

- [x] test scripts updated


JoernBerkefeld:
here is the change i pushed: https://github.com/Accenture/sfmc-devtools/pull/2304/files

image

Very good that you identified config.getProperties() :) 
what i added on top was checking Util.OPTIONS.format - which allows you to toggle formatting on-the-fly.

Looking forward to your next PR! Hope I will not miss that as well :(

PR #2370, opened 2025-09-28
Feature/2368 minor improvements to filter optional parameter functionality

### LINKED ISSUE #1234: [TASK] investigate minimum scope / permissions required to run mcdev

phjulia:
# PR details

2 minor changes to improve experience when using --filter option with retrieve command. Previously, the user couldn't see which items the function filtered out, also the command wiped the folder as if it were to re-retrieve all items.

Logging level changed, and expanded the condition in the if statement. 

- closes #1234

## Further details (optional)

...

## Checklist

- [N/A ] test scripts updated
- [N/A] Wiki updated (if applicable)


lib/index.js
```diff
@@ -436,7 +436,7 @@ class Mcdev {
                     const areKeySet = Array.isArray(selectedTypesArr)
                         ? !!keys
                         : selectedTypesArr[selectedType] !== null;
-                    if (!areKeySet && !Util.OPTIONS.like) {
+                    if (!areKeySet && Util.OPTIONS.purge) {
```
JoernBerkefeld: this should be an OR rather than an AND.
if we dont set keys, we always want that folder to be hard-resetted

PR #2372, opened 2025-09-29
Feature/2371 feature add new metadata type triggeredsendsummary

### LINKED ISSUE #1234: [TASK] investigate minimum scope / permissions required to run mcdev

phjulia:
# PR details
Added new type triggeredSendSummary

Added tests, metadata type files for the new type

- closes #1234

## Further details (optional)

...

## Checklist

- [ ] test scripts updated


lib/metadataTypes/TriggeredSendSummary.js
```diff
@@ -0,0 +1,100 @@
...
+            const triggeredSendKey = cache.searchForField(
+                'triggeredSend',
+                metadata.CustomerKey,
+                'CustomerKey',
+                'CustomerKey'
+            );
+            metadata.r__triggeredSend_key = triggeredSendKey;
+        } catch {
+            // the function will throw an error if the field is not found. We don't need to store the summary if the related TriggeredSend is not found
+            throw new Error(
```
JoernBerkefeld: dont throw an error in postRetrieveTasks to skip the item but instead ensure the function returns null or undefined. That will then be caught one level up and the item will be recorded as "filtered".

image

any logs generated by postRetrieveTasks usually use Util.logger.warn or Util.logger.verbose()
verbose is only visible in the log-file while warn is used for cases you really need the user to see. My guess is, verbose is the better option here.

lib/metadataTypes/definitions/TriggeredSendSummary.definition.js
```diff
@@ -0,0 +1,219 @@
...
+            isUpdateable: false,
+            retrieving: false /** Listed in the type documentation but not retrievable */,
+            templating: false,
+        },
+        'TriggeredSend.CategoryID': {
+            isCreateable: false,
+            isUpdateable: false,
+            retrieving: false,
+            templating: false,
+        },
```
JoernBerkefeld: do you really want to see that in the final json or did you just intend to use it temporarily to make `this.setFolderPath` work? I would remove that after the folder was searched for as the id itself adds no value - and then you dont need to mention it here.

in my test retrieval i got a lot of items looking like this:

image

lib/metadataTypes/TriggeredSendSummary.js
```diff
@@ -0,0 +1,100 @@
...
+            // folder path for a TriggeredSendSummary is the same as for the related TriggeredSend. CategoryID is not retrievable for this type, so use the TriggeredSend's field
+            this.setFolderPath(metadata);
+            const triggeredSendName = cache.searchForField(
+                'triggeredSend',
+                metadata.CustomerKey,
+                'CustomerKey',
+                'Name'
+            );
+            metadata.r__triggeredSend_name = triggeredSendName;
+            const triggeredSendKey = cache.searchForField(
```
JoernBerkefeld: you probably want to use cache.getByKey(type, key) first, and only if that returned the TS object then set the other values. 
Your current approach basically searches for the same object 3 times (to get the folder id, the name and confirm the key).

Therefore, I would suggest you rework this entire method based on the above suggestion which could then in fact also lead to only one log entry

JoernBerkefeld (changes requested):

lib/metadataTypes/definitions/TriggeredSendSummary.definition.js
JoernBerkefeld: from the log file for the above example:
image

setFolderPath issued a warning for this but by design does not remove the id in these moments. We, however, should manually remove it after setFolderPath ran to ensure it doesnt show up in the json. One could argue that if the folder no longer exists then potentially the summary also makes no sense but i dont understand the functionality of tsSummary enough to make that judgement call

yuliialikhyt:
@JoernBerkefeld Thank you :) I agree with your comments. I think I addressed them all.

lib/metadataTypes/definitions/TriggeredSendSummary.definition.js
```diff
@@ -0,0 +1,213 @@
...
+    dependencyGraph: {
+        triggeredSend: ['r__triggeredSend_key'],
+    },
+    filter: {
+        r__folder_Path: [
+            'Journey Builder Sends/',
+            'HiddenCategory/HiddenJourneyBuilderTriggeredSends/',
+        ],
+    },
+    folderType: 'triggered_send_summary',
```
JoernBerkefeld: thats used for creating records only, which you dont do. remove

lib/metadataTypes/definitions/TriggeredSendSummary.definition.js
```diff
@@ -0,0 +1,213 @@
...
+    hasExtended: false,
+    idField: 'ObjectID',
+    keepId: true,
```
JoernBerkefeld: thats used to keep the id field in the json. set to false

lib/metadataTypes/definitions/TriggeredSendSummary.definition.js
```diff
@@ -0,0 +1,213 @@
...
+    keyIsFixed: true,
+    keyField: 'CustomerKey',
+    nameField: 'Name',
```
JoernBerkefeld: this should point to r__triggeredSend_name

lib/metadataTypes/TriggeredSendSummary.js
```diff
@@ -0,0 +1,92 @@
...
+        metadata['TriggeredSend.CategoryID'] = triggeredSend.CategoryID;
+        // folder path for a TriggeredSendSummary is the same as for the related TriggeredSend. CategoryID is not retrievable for this type, so use the TriggeredSend's field
+        this.setFolderPath(metadata);
+        if (!metadata.r__folder_Path) {
+            Util.logger.verbose(
+                ` ☇ skipping ${this.definition.typeName} '${metadata.Name}'/'${metadata.CustomerKey}': Could not find folder.`
+            );
+            // do not save this Triggered Send Summary because it would not be visible in the user interface
+            return;
+        }
```
JoernBerkefeld, line 94:
```suggestion
        try {
            metadata.r__folder_Path = cache.searchForField(
                'folder',
                metadata[this.definition.folderIdField],
                'ID',
                'Path'
            );
            delete metadata[this.definition.folderIdField];
        } catch (ex) {
            Util.logger.verbose(
                ` ☇ skipping ${this.definition.typeName} '${metadata.Name}'/'${metadata.CustomerKey}': Could not find folder.`
            );
            // do not save this Triggered Send Summary because it would not be visible in the user interface
            return;
        }

```

i believe you need a custom version of setFolderPath here or else your log gets blasted with warnings (happend during my test).

ideally you actually create a clone of setFolderPath in your class to keep it clean. the above would be the base logic

lib/metadataTypes/TriggeredSendSummary.js
```diff
@@ -0,0 +1,92 @@
...
+            // move original filter down one level into rightOperand and add key filter into leftOperand
+            requestParams = {
+                filter: {
+                    leftOperand: {
+                        leftOperand: 'CustomerKey',
+                        operator: 'equals',
+                        rightOperand: key,
+                    },
+                    operator: 'AND',
+                    rightOperand: requestParams.filter,
```
JoernBerkefeld: we dont have another filter in this request as far as i can see. i would therefore ask to remove the outer left/right operand object and only keep the inner one that filters on CustomerKey

lib/metadataTypes/definitions/TriggeredSendSummary.definition.js
```diff
@@ -0,0 +1,213 @@
+export default {
+    bodyIteratorField: 'Results',
+    dependencies: ['triggeredSend', 'folder'],
```
JoernBerkefeld:
```suggestion
    dependencies: [
        'triggeredSend',
        'folder-hidden',
        'folder-triggered_send',
        'folder-triggered_send_journeybuilder',
    ],
```

copied this from triggeredSend - limits the amount of folders that get retrieved to only relevant subtypes and hence speeds up retrieval.

lib/metadataTypes/definitions/TriggeredSendSummary.definition.js
```diff
@@ -0,0 +1,213 @@
...
+    dependencyGraph: {
+        triggeredSend: ['r__triggeredSend_key'],
+    },
+    filter: {
+        r__folder_Path: [
+            'Journey Builder Sends/',
+            'HiddenCategory/HiddenJourneyBuilderTriggeredSends/',
+        ],
```
JoernBerkefeld:
```suggestion
```

remove - this is whats removing your summary records. I dont want those triggered sends to be downloaded but if you want to be able to find any summaries, even for journeys, then thats what you need removed.

lib/metadataTypes/definitions/TriggeredSendSummary.definition.js
```diff
@@ -0,0 +1,213 @@
...
+    keyField: 'CustomerKey',
+    nameField: 'Name',
+    folderIdField: 'TriggeredSend.CategoryID',
+    createdDateField: null,
+    createdNameField: null,
+    lastmodDateField: null,
+    lastmodNameField: null,
+    restPagination: null,
+    maxKeyLength:
+        null /** Key can't be changed, it's the same as the TriggeredSend key, metadata of this type can't be created either */,
```
JoernBerkefeld:
```suggestion
    maxKeyLength: 36,
```

copied from triggeredSend.definition

JoernBerkefeld (changes requested):
you will probably have to re-do the test-data

PR #2420, opened 2025-11-13
feature/2252 resolve Update Contact Data activities in Journeys

### LINKED ISSUE #2252: [FEATURE] resolve Update Contact Data activities in Journeys
- [ ] UPDATECONTACTDATAV2-1
Image

--> dataExtension + dataExtensionField

yuliialikhyt:
# PR details

Before
image

After
image

Added some changes to have the Data Extension field name resolved and add the Data Extension key to the journey instead of ObjectID. 

- closes #2252 

## Further details (optional)

...

## Checklist

- [x] test scripts updated


lib/metadataTypes/definitions/Journey.definition.js
```diff
@@ -8,7 +8,8 @@ export default {
     dependencies: [
         'folder-journey',
         'triggeredSend', // for EMAILV2-activity
-        'dataExtension', // for transactionalEmails: EMAILV2-activity
+        'dataExtension', // for transactionalEmails: EMAILV2-activity and for UPDATECONTACTDATA activity
```
JoernBerkefeld, line 11: Good, please also update dependencyGraph section down below (can't comment directly...) lines 28ff

lib/metadataTypes/definitions/Journey.definition.js
```diff
@@ -716,6 +717,18 @@ export default {
         'activities[].schema': {
             skipValidation: true,
         },
+        'activities[].arguments.activityData.updateContactFields[].r__dataExtension_key': {
+            isCreateable: true,
+            isUpdateable: true,
```
JoernBerkefeld: Should not be creatable nor updatable because it's a custom field and hence may not be send to the soap endpoint

test/type.journey.test.js
JoernBerkefeld, line 1: Would be nice to see a single test case for this, maybe baked into the existing ones?

lib/metadataTypes/definitions/Journey.definition.js
```diff
@@ -8,7 +8,8 @@ export default {
     dependencies: [
         'folder-journey',
         'triggeredSend', // for EMAILV2-activity
-        'dataExtension', // for transactionalEmails: EMAILV2-activity
+        'dataExtension', // for transactionalEmails: EMAILV2-activity and for UPDATECONTACTDATA activity
+        'dataExtensionField', // for UPDATECONTACTDATA activity
```
JoernBerkefeld: I wonder if this dependency is a good idea here. On some BUs, due to bad past decisions, there might be 50k DEs with 500k fields. That takes forever to cache and won't always be needed. Maybe the fields could be cached when the need arises only? We are doing ad-hoc caching in some places already. I think for events we are doing that to cover salesforce triggered journeys

lib/metadataTypes/DataExtensionField.js
```diff
@@ -47,11 +46,7 @@ class DataExtensionField extends MetadataType {
      * @returns {Promise.<MetadataTypeMapObj>} Promise of items
      */
     static async retrieveForCache() {
-        const cachedDEs = cache.getCache().dataExtension;
-        if (cachedDEs) {
-            await DataExtension.attachFields(cachedDEs);
```
JoernBerkefeld, line 52: Can't help but wonder why this was added here. Will dig into file history a bit to figure this out. There must have been a reason at some point

lib/metadataTypes/Journey.js
```diff
@@ -1239,6 +1239,49 @@ class Journey extends MetadataType {
                     }
                     break;
                 }
+                case 'UPDATECONTACTDATA': {
+                    const activityArguments = activity.arguments;
+                    for (
+                        let i = 0;
+                        i < activityArguments.activityData.updateContactFields.length;
```
JoernBerkefeld: You are assuming activityData and updateContactFields both exist here. Better to use ?. notation to make sure you dont trigger a js error

lib/metadataTypes/Journey.js
```diff
@@ -1239,6 +1239,49 @@ class Journey extends MetadataType {
...
+                        i++
+                    ) {
+                        const contactField = activityArguments.activityData.updateContactFields[i];
+                        try {
+                            activityArguments.activityData.updateContactFields[
```
JoernBerkefeld: Use contactField var instead

lib/metadataTypes/Journey.js
```diff
@@ -1239,6 +1239,49 @@ class Journey extends MetadataType {
...
+                                    'Name'
+                                );
+                        } catch (ex) {
+                            Util.logger.warn(
+                                ` - ${this.definition.type} '${metadata[this.definition.nameField]}' (${
+                                    metadata[this.definition.keyField]
+                                }) activity-key=${activity.key}: ${ex.message}`
+                            );
+                        }
+                    }
```
JoernBerkefeld, line 1289: Ok, that's handling postRetrieve. Please also add reverse logic in preDeploy

test/type.journey.test.js
yuliialikhyt: Added two test separate cases, one for deploy and one for retrieve

lib/metadataTypes/Journey.js
yuliialikhyt: Added logic into preDeploy

lib/metadataTypes/definitions/Journey.definition.js
yuliialikhyt: Removed the dependency, and added retrieve of DataExtensionField in the case statement itself (preDeploy and postRetrieve)

lib/metadataTypes/definitions/Journey.definition.js
yuliialikhyt: Updated

lib/metadataTypes/definitions/Journey.definition.js
yuliialikhyt: I've added 'field' too, it was there by default, but I'm changing the value from ObjectID to Name. Not sure if I should keep it? I added a new field and tested deploy  of the journey - it did not cause any problems. 
```
        'activities[].arguments.activityData.updateContactFields[].field': {
            isCreateable: false,
            isUpdateable: false,
            retrieving: true,
            template: true,
        },
```

lib/metadataTypes/Journey.js
yuliialikhyt: Updated

lib/metadataTypes/Journey.js
yuliialikhyt: updated

lib/metadataTypes/Journey.js
```diff
@@ -1239,6 +1240,53 @@ class Journey extends MetadataType {
...
+                            );
+                        }
+                        // Don't replace the field Id with the name if DE CustomerKey was not found
+                        if (contactField.r__dataExtension_key) {
+                            Util.logger.warn(' - Caching dependent Metadata: dataExtensionField');
+                            DataExtensionField.buObject = this.buObject;
+                            DataExtensionField.client = this.client;
+                            DataExtensionField.properties = this.properties;
+                            const dataExtensionFieldCache =
+                                await DataExtensionField.retrieveForCache();
```
JoernBerkefeld: Only re-cache if dataExtensionFields were not already cached please. where you placed it, this caching gets executed once for every journey in your retrieve

You can find an example of how i did that in Journey.deploy and Journey._postRetrieveTasksBulk:

image

image

I would recommend putting YOUR caching also into those 2 methods in a similar fashion (quick loop throuh journeys and the activities to check if the activity is even in the retrieve/deploy payload

lib/metadataTypes/Journey.js
```diff
@@ -1239,6 +1240,53 @@ class Journey extends MetadataType {
                     }
                     break;
                 }
+                case 'UPDATECONTACTDATA': {
+                    const contactFields = activity.arguments?.activityData?.updateContactFields;
+                    for (const contactField of contactFields) {
```
JoernBerkefeld, line 1247: missing an if-check on contactFields being an Array first or else this could cause a js error

lib/metadataTypes/Journey.js
```diff
@@ -1239,6 +1240,53 @@ class Journey extends MetadataType {
...
+                        } catch (ex) {
+                            Util.logger.warn(
+                                ` - ${this.definition.type} '${metadata[this.definition.nameField]}' (${
+                                    metadata[this.definition.keyField]
+                                }) activity-key=${activity.key}: ${ex.message}`
+                            );
+                        }
+                        // Don't replace the field Id with the name if DE CustomerKey was not found
+                        if (contactField.r__dataExtension_key) {
+                            Util.logger.warn(' - Caching dependent Metadata: dataExtensionField');
```
JoernBerkefeld:
```suggestion
                            Util.logger.info(' - Caching dependent Metadata: dataExtensionField');
```

lib/metadataTypes/Journey.js
```diff
@@ -1239,6 +1240,53 @@ class Journey extends MetadataType {
...
+                            DataExtensionField.client = this.client;
+                            DataExtensionField.properties = this.properties;
+                            const dataExtensionFieldCache =
+                                await DataExtensionField.retrieveForCache();
+                            cache.setMetadata(
+                                'dataExtensionField',
+                                dataExtensionFieldCache.metadata
+                            );
+                            try {
+                                contactField.field = cache.searchForField(
```
JoernBerkefeld: i get why you wanted to stick with "field" but the standard here would be `r__dataExtensionField_name`

lib/metadataTypes/Journey.js
```diff
@@ -1666,6 +1714,40 @@ class Journey extends MetadataType {
...
+                            Util.logger.warn(
+                                ` - ${this.definition.type} '${metadata[this.definition.nameField]}' (${
+                                    metadata[this.definition.keyField]
+                                }) activity-key=${activity.key}: ${ex.message}`
+                            );
+                        }
+                        try {
+                            contactField.field = cache.searchForField(
+                                'dataExtensionField',
+                                `[${contactField.r__dataExtension_key}].[${contactField.field}]`,
```
JoernBerkefeld:
```suggestion
                                `[${contactField.r__dataExtension_key}].[${contactField.r__dataExtensionField_name}]`,
```

lib/metadataTypes/definitions/Journey.definition.js
```diff
@@ -716,6 +717,18 @@ export default {
...
         },
+        'activities[].arguments.activityData.updateContactFields[].r__dataExtension_key': {
+            isCreateable: false,
+            isUpdateable: false,
+            retrieving: true,
+            template: true,
+        },
+        'activities[].arguments.activityData.updateContactFields[].field': {
+            isCreateable: false,
+            isUpdateable: false,
```
JoernBerkefeld, line 728:
```suggestion
            isCreateable: true,
            isUpdateable: true,
```
this one you DO want creatable and updatable because it's a standard field.

please also add the new custom field i suggested (r__dataExtensionField_name)

lib/metadataTypes/Journey.js
JoernBerkefeld: just realized, there might be a slightly better option here:
u can use something like in **DataExtension._retrieveFieldsForSingleDe()** - that way it works nicely even on BUs with a extremely large amount of DEs / fields.

that said, this will mean quite a bit of extra work

lib/metadataTypes/Journey.js
yuliialikhyt: Sorry, I forgot to push some commits 
updateContactFields can be either undefined or an array, so I think this should work? (looks shorter)
 const contactFields = activity?.arguments?.activityData?.updateContactFields ?? [];

lib/metadataTypes/Journey.js
yuliialikhyt: I've added a function in the DataExtensionField like the one in the DataExtension `retrieveFieldsForSingleDe`. Let me know if that is ok 🙂

lib/metadataTypes/Journey.js
```diff
@@ -1666,6 +1713,51 @@ class Journey extends MetadataType {
...
+                        }
+                        try {
+                            if (!cache.getCache().dataExtensionField) {
+                                DataExtensionField.buObject = this.buObject;
+                                DataExtensionField.client = this.client;
+                                DataExtensionField.properties = this.properties;
+                                const fields = await DataExtensionField.retrieveFieldsForSingleDe(
+                                    contactField.r__dataExtension_key
+                                );
+                                cache.setMetadata('dataExtensionField', fields);
```
JoernBerkefeld, line 1750: this wont work because now your cache for dataExtensionField is existing - but you only inserted fields for a single DE. Therefore, if you run this code for 2 updatecontactdata activities with a different dataExtension then the second one will not recache fields and hence fail to find any.

JoernBerkefeld: what could work is if you were to reset the dataExtensionField cache after running searchForField again - but only if you didnt find a dataExtensionField cache in the first place? bit hacky but should be most efficient that way.

```js
let tempCachedFields = false;
if (!cache.getCache().dataExtensionField) {
    tempCachedFields = true;
    ...
}
...  cache.searchForField(..);
if(tempCachedFields) {
    cache.clearCache(this.buObject.mid, 'dataExtensionField');
}
```

lib/metadataTypes/Journey.js
```diff
@@ -1239,6 +1240,52 @@ class Journey extends MetadataType {
...
+                            }
+                            contactField.r__dataExtensionField_name = cache.searchForField(
+                                'dataExtensionField',
+                                contactField.field,
+                                'ObjectID',
+                                'Name'
+                            );
```
JoernBerkefeld, line 1280: see comment on resetting the cache in `_preDeployTasks_activities` and solve similarly here

lib/metadataTypes/DataExtensionField.js
```diff
@@ -47,11 +46,28 @@ class DataExtensionField extends MetadataType {
...
-        }
-        return;
```
JoernBerkefeld, line 54: please revert as you are not using this change. i believe whenever we want to cache fields, we ended up looking at them through dataExtensions alone and hence had the above logic in place

lib/metadataTypes/DataExtensionField.js
```diff
@@ -426,7 +442,6 @@ class DataExtensionField extends MetadataType {
 
 // Assign definition to static attributes
 import MetadataTypeDefinitions from '../MetadataTypeDefinitions.js';
-import cache from '../util/cache.js';
```
JoernBerkefeld, line 429: please restore in favor of old version of `retrieveForCache()`

JoernBerkefeld (changes requested):

lib/metadataTypes/DataExtensionField.js
yuliialikhyt: reverted retrieveForCache() and brought back the imports

lib/metadataTypes/Journey.js
yuliialikhyt: Yeap, makes sense. I made the changes as suggested, in preDeploy and postRetrieve. Since the function retrieves all fields per DE, and one Update Contact activity is related to one DE only, cache is cleared once per UPDATECONTACTDATA activity. It will not re-cache once for every field, but once per every data extension / Update Contact activity.

lib/metadataTypes/Journey.js
```diff
@@ -1239,6 +1240,58 @@ class Journey extends MetadataType {
                     }
                     break;
                 }
+                case 'UPDATECONTACTDATA': {
```
JoernBerkefeld, line 1243: im gonna approve this but i do have one remaining question id love to talk about (aside from this PR):
whats the difference between this type `SALESCLOUDACTIVITY` (which I'm using in some of my journeys) and `UPDATECONTACTDATA`?

yuliialikhyt: UPDATECONTACTDATA is the "Update Contact" JB activity, it updates the contact record in the data extension. We use it for logging where the contact is in the journey. 
SALESCLOUDACTIVITY is a group of activities that create or update contacts/accounts/other objects in the connected Sales Cloud instance.

PR #2421, opened 2025-11-14
Bug/2337 journey created date is incorrect in update response

### LINKED ISSUE #2337: [BUG] journey createdDate incorrect in update-response
basically, we should override the response from the update-response for the field `createdDate` with what was returned while caching the journeys. otherwise it looks like this:

Image

yuliialikhyt:
# PR details

Updated the createdDate in the journey with the cached value in post deploy. 

- closes #2337

## Further details (optional)

...

## Checklist

- [ ] test scripts updated


lib/metadataTypes/Journey.js
```diff
@@ -1880,6 +1880,10 @@ class Journey extends MetadataType {
      * @param {MetadataTypeMap} upsertResults metadata mapped by their keyField as returned by update/create
      */
     static async postDeployTasks(upsertResults) {
+        // override the response from update by createdDate from cache (the date from update-response is incorrect)
+        for (const [key, journey] of Object.entries(upsertResults)) {
+            journey.createdDate = cache.getByKey(this.definition.type, key)?.createdDate;
```
JoernBerkefeld: That doesn't seem to check if there is a match in the cache and consequently resets this to empty on newly created entries.

yuliialikhyt: Updated

PR #2528, opened 2026-03-03
Feature/2399-share-assets

### LINKED ISSUE #2399: [FEATURE] Automatic Asset Sharing After Deployment
**Is your feature request related to a problem? Please describe.**
In our project, we deploy market-specific assets that are later reused in templates. Currently, after each deployment, we have to manually share these assets with the corresponding business unit. This manual process is time-consuming and repetive

**Describe the solution you'd like**
I'd like the Devtools to automatically share an asset with target bussiness unit based on the `sharingProperties` parameter from assets

**Describe alternatives you've considered**
- New post-deployment "share" task that shares selected assets based on the `sharingProperties` parameter from asset
- Adding `--share` flag to the `deploy` command, which triggers automatic sharing using the `sharingProperties` parameter from assets

kacperkruger:
# PR details

- Added option to automaticlly share assets during deployment.

- Added `--shareAsset` option for building templates and deploying components

- Changed stringification fields logic before templating to support nested paths with wildcards

- closes #2399

## Further details (optional)

...

## Checklist

- [x] test scripts updated


lib/metadataTypes/MetadataType.js
```diff
@@ -410,21 +410,48 @@ class MetadataType {
         if (this.definition.stringifyFieldsBeforeTemplate) {
             // numeric fields are returned as numbers by the SDK/API. If we try to replace them in buildTemplate it would break the JSON format - but not if we stringify them first because then the {{{var}}} is in quotes
             const metadataTemp = JSON.parse(metadataStr);
-            for (const field of this.definition.stringifyFieldsBeforeTemplate) {
```
JoernBerkefeld, line 413: Can you please explain why this change was necessary?

kacperkruger: Fields containing business unit id as Integer (`businessUnitAvailability.%.memberId`, `sharingProperties.sharedWith`) have to be stringified before templating. Neded to add support for nested paths and wildcards paths

kacperkruger: Let me know if you think it should be refactored or done differently

JoernBerkefeld: Actually like it. Wondering if we don't already have that same logic as part of how type dependencies are tracked for build --dependencies. If I am right I'd like to move the logic somewhere central and reuse it. Will investigate that

JoernBerkefeld:
@copilot pls fix

kacperkruger:
Hi @JoernBerkefeld, do you have any estimate for when this feature might be released? Also please let me know if there's anything else needed from my side to move this forward

JoernBerkefeld:
Hi @kacperkruger - sorry about the delay. I've been on parental leave and concentrated on my privately published sfmc tools these pasts months due to it.

Let me prioritize your PR this week.

lib/cli.js
```diff
@@ -534,6 +538,10 @@ yargs(hideBin(process.argv))
                 .option('fix', {
                     group: 'Options for build:',
                     describe: 'auto-fix validation issues if the rule is able to do it',
+                })
+                .option('shareAsset', {
+                    group: 'Options for buildTemplate:',
```
JoernBerkefeld:
```suggestion
                    group: 'Options for build:',
```

JoernBerkefeld:
heads-up: im rewriting this PR a bit to fit better into the release. intented outcomes stays. just the how changes

JoernBerkefeld (approved):

JoernBerkefeld:
I found that only templating was removing that sharing info in the first place, specifically, the buildtemplate functionality.

Therefore, I removed your new flag from other functions that would not use it anyways.

Looking good so far from my end.

PR #2645, opened 2026-09-11
function/2644: asset-email & v10 asset migration for cdp

### LINKED ISSUE #2644: [FEATURE] regroup and migrate asset subtypes
asset-message should become asset-email.
we have already regrouped quite a bit of asset subtypes for the coming release to asset-mobile and asset-webstudio and hence it would be only fitting to also change asset-message now rather than later.

doing so will likely cause a lot of false-positives in cdp and therefore we should handle that somehow

JoernBerkefeld:
# PR details

- closes #2644

## Further details (optional)

...

## Checklist

- [ ] test scripts updated


JoernBerkefeld:
re-doing this...

